PR #4, opened 2019-02-20
Rename maven artifact to 'com.ibm.cloud:sdk-core'

padamstx:


pom.xml
```diff
@@ -2,11 +2,11 @@
...
     <modelVersion>4.0.0</modelVersion>
-    <groupId>com.ibm.cloud.sdk</groupId>
-    <artifactId>java-sdk-core</artifactId>
+    <groupId>com.ibm.cloud</groupId>
+    <artifactId>sdk-core</artifactId>
     <packaging>jar</packaging>
-    <name>java-sdk-core</name>
-    <version>1.0-SNAPSHOT</version>
+    <name>sdk-core</name>
+    <version>99-SNAPSHOT</version>
```
lpatino10: This doesn't really matter since you set it in the Travis script, right?

padamstx: correct, 99-SNAPSHOT would just be the default artifact version # for a non-tagged build

PR #16, opened 2019-03-14
Replace reactive programming library

lpatino10:
Related to https://github.com/watson-developer-cloud/java-sdk/issues/1013

This PR replaces the current reactive programming library (https://mvnrepository.com/artifact/org.glassfish.jersey.bundles.repackaged/jersey-jsr166e) with [RxJava](https://github.com/ReactiveX/RxJava).

This library was chosen because it is:
- popular among Java developers
- used in high-profile projects
- currently maintained/updated
- Java 7-compatible

Tests have been added in this PR to ensure the two new execution method (`reactiveRequest()`) work s properly, while documentation has been updated and is currently in review in the Watson SDK repo: https://github.com/watson-developer-cloud/java-sdk/pull/1045

**Along with the above changes,** both `executeWithDetails()` and `enqueueWithDetails()` have been removed. The normal version of those methods now return the same wrapped HTTP response object, so that we always return the IBM Cloud API response + extra info.


germanattanasio:
I think we want to default to detailed responses so maybe one of the reactive responses can be eliminated.

lpatino10:
Ah yeah I forgot about that actually. I'll go ahead and do that in this PR then for all versions.

PR #17, opened 2019-03-15
Make IAM API call on separate thread

lpatino10:
Resolves [this issue](https://github.com/watson-developer-cloud/java-sdk/issues/1041).

This change ensures that users don't have to make any special changes on Android going from U/P auth to IAM.


src/main/java/com/ibm/cloud/sdk/core/service/security/IamTokenManager.java
```diff
@@ -181,14 +183,32 @@ private boolean isRefreshTokenExpired() {
    * @return object containing requested IAM token information
    */
   private IamToken callIamApi(Request request) {
-    Call call = HttpClientSingleton.getInstance().createHttpClient().newCall(request);
-    ResponseConverter<IamToken> converter = ResponseConverterUtils.getObject(IamToken.class);
+    final IamToken[] returnToken = new IamToken[1];
```
germanattanasio: Does this work in Android? I think their have their own Thread management system

lpatino10: Just managed to test it out and it looks like Android accepts the Java threading just fine 👌

PR #19, opened 2019-03-18
Revert old SSL/TLS logic change

lpatino10:
A while ago, a change was made to the Java SDK core to try and fix some `SSLHandshakeException`s showing up in Visual Recognition on Oracle JDK 7. The PR to try and make that fix is here: https://github.com/watson-developer-cloud/java-sdk/pull/890

However, there was an issue reported later that this "fix" introduced some other potential problems. Here is that issue: https://github.com/watson-developer-cloud/java-sdk/issues/1020

I went back to investigate by using the old logic and Oracle JDK 7 with Visual Recognition and I didn't run into any problems. Based on the timing of the original occurrence, my guess is it had something to do with the transition to IAM authentication and the new endpoint and was just temporary. Either way, if the original logic works and avoids the potential problems that were reported, I think that's preferred.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
@@ -186,8 +186,8 @@ public class HttpClientSingleton {
 
       X509TrustManager trustManager = (X509TrustManager) trustManagers[0];
 
-      System.setProperty("com.ibm.jsse2.overrideDefaultTLS", "true");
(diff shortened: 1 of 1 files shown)


germanattanasio:
My only concern is if this will work with the IBM JDK rather than the Oracle JDK.

You could pull a docker image with the IBM JDK and try to execute this there.

germanattanasio:
Looks like it was fixed in https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-370

PR #20, opened 2019-03-25
Only include nulls in serialization when processing API responses

lpatino10:
In [this earlier PR](https://github.com/IBM/java-sdk-core/pull/18), I changed the main GSON instance to always serialize with null values. When testing with the new core, I realized that some API endpoints don't like being sent null values. Originally the change was only made with service responses in mind, so I've tweaked the core logic accordingly. JSON content being sent to the APIs should now behave as before.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -246,8 +246,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${compiler-plugin-version}</version>
                 <configuration>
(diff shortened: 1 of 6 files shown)


lpatino10:
I've also made this change on the `v2` branch, so if this is approved, I can tag and release that as a fix for anyone using the `2.0.1` version.

PR #22, opened 2019-04-09
Credential file IAM fix

lpatino10:
Related to https://github.ibm.com/Watson/developer-experience/issues/6452

Basically, we were previously using the name `apikey` for IAM API keys. This was a name that our team didn't really want to use on IBM Cloud because it might be ambiguous, but it's what stuck. Regardless, we've chosen to use `iam_apikey` in our own code, and so the library to download the credential file was changed accordingly. This change just follows suit to still work with that update.

We do still check for the old `apikey` name so as not to break anyone who may have downloaded a credential file previously.

Lastly, I also included a change to try and pull the error message from bad IAM responses, which use the key `errorMessage`. This is a small change just to try and make those problems more readable.


lpatino10:
**Not** ready to merge until the change is present in the service dashboard. See https://github.ibm.com/Watson/service-dashboard-pages/pull/375

PR #27, opened 2019-04-24
Add release docs to GH pages site on releases

lpatino10:
This PR pulls from very similar work that was done in the Watson Java SDK to publish the Javadoc on releases. I've already added the site URL to the README, and you can find it to reference here: https://ibm.github.io/java-sdk-core/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,11 @@ deploy:
     skip_cleanup: true
     on:
       tags: true
+  - provider: script
+    script: ".build/push-javadoc-to-gh-pages.sh"
+    skip_cleanup: true
+    on:
+      tags: true
 
 # Ask bintray to sync the just-deployed artifact to maven central
 after_script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -54,6 +54,11 @@ deploy:
     skip_cleanup: true
     on:
       tags: true
+  - provider: script
+    script: ".build/push-javadoc-to-gh-pages.sh"
```
padamstx: Shouldn't this be "./build..."?

build/push-javadoc-to-gh-pages.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" ]; then
+
+  git config --global user.email "wps@us.ibm.com"
```
padamstx: Ummm, do we have to use a "watson"-related name here?  :)

lpatino10: The dangers of copy and paste. I don't think I need those user lines anyway 😄

PR #28, opened 2019-04-25
feat(IAM): set client id/secret via IamOptions; corrected terminology

padamstx:
This PR fixes the initial changes for issue https://github.ibm.com/arf/planning-sdk-squad/issues/851 :
1. The client id/secret can now be set via the IamOptions class.   This will allow SDK users to configure these values as part of instantiating the generated service object (i.e. the ctor accepts an IamOptions instance).
2. I changed "secret" to be "clientSecret" to be more descriptive and to align with "clientId".
3. The client id/secret values are no longer static.

The actual requirement is to allow SDK users to set the client id/secret values so that interactions with the IAM token server will use an Authorization header that reflects their specific id/secret rather than the default of "bx:bx".

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/security/IamOptions.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/security/IamOptions.java
@@ -19,6 +19,9 @@ public class IamOptions {
   private String apiKey;
   private String accessToken;
   private String url;
+  private String clientId;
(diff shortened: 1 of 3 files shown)


padamstx:
@lpatino10 had previously approved this PR, but that approval was dismissed after I force-pushed after rebasing my commit on the updated master branch.

PR #29, opened 2019-04-27
Fix Javadoc deployment

lpatino10:
This PR fixes my last one to make sure generating the Javadoc and pushing them to the `gh-pages` branch actually works. The changes I had to make were:
- Calling the `mvn javadoc:javadoc` command to actually generate the docs
- Modifying the new scripts to allow them to be executed
- Fixing the path calling the `generate-index-html` script


.travis.yml
```diff
@@ -37,6 +37,7 @@ before_script:
 
 script:
 - mvn clean package
+- mvn javadoc:javadoc
```
padamstx: This additional command should work fine, but I think if you had changed the `package` goal in the first command to be `install` that it would also create the javadocs (I think... it might also require an additional pom.xml config change as well, not 100% sure)).   Just for future reference :)

.travis.yml
```diff
@@ -57,8 +58,6 @@ deploy:
   - provider: script
     script: "./build/push-javadoc-to-gh-pages.sh"
     skip_cleanup: true
-    on:
```
padamstx: Hmmm, I don't think we want to deploy on all builds, do we?   That's why I had set the `on: tags` qualifier.

.travis.yml
lpatino10: Oh, probably not then 😕  I wasn't sure originally if that was the issue we were having.

.travis.yml
germanattanasio: The script was designed to deploy branches too
See: http://watson-developer-cloud.github.io/java-sdk/

PR #30, opened 2019-04-29
feat: add de-serialization logic for JSON arrays within response body

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/797 (for Java)

This PR adds the necessary function (and tests) to the Java SDK core to allow us to generate Java code that will support the de-serialization of a top-level (JSON) array within a response body.
With this change, the Java SDK core will support responses which contain an array of JSON object as well as primitive types.
Note that this PR will be a prereq to some related changes in the SDK generator.  That PR will be finished soon.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/ResponseConverterUtils.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/ResponseConverterUtils.java
@@ -65,8 +65,7 @@ public final class ResponseConverterUtils {
   }
 
   /**
-   * Creates a generic {@link ResponseConverter} for a POJO class. <br>
-   * It should extend {@link ObjectModel}
+   * Creates a generic {@link ResponseConverter} for a POJO class that extends ObjectModel.
    *
    * @param <T> the generic type
    * @param type the type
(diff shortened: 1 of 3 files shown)


germanattanasio:
Do we want to support this? I thought that top-level arrays were against our [guidelines](https://github.com/watson-developer-cloud/api-guidelines#anonymous-json-arrays)

padamstx:
Ha! I thought I had added this reply to German's question earlier but apparently got distracted before I could click `Comment` :)

> Do we want to support this? I thought that top-level arrays were against our [guidelines](https://github.com/watson-developer-cloud/api-guidelines#anonymous-json-arrays)

Yes, while it's prohibited by the API guidelines, we still need to support the feature due to a requirement from one of our other customers where they need to maintain compatibility with a third-party API definition (e.g. the internal Cloudant team needs to maintain support for the CouchDB API).

germanattanasio:
I assumed that was the case and I'm familiar with the CouchDB API. It makes sense to try to mimic their API so that we can help developers familiar with it but interested in Cloudant 👍

padamstx:
I'll hold off on merging this PR in until the Java code generator changes are finished.

PR #35, opened 2019-05-08
feat(http): improve SDK's ability to configure http client instances

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/859

This PR allows a Java SDK to be more easily modify the configuration of the `OkHttpClient` instance used by the Java core for REST API invocations.   With these changes, an SDK user would also have the ability to create and set their own `OkHttpClient` instance for use by the SDK (similar to the Go core).

The new tests added to the BaseServiceTest class should demonstrate the gist of these changes.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
@@ -87,10 +87,14 @@ public class HttpClientSingleton {
     return instance;
   }
 
+  /**
+   * All new OkHttpClient instances are created from this instance, which contains
+   * a default configuration.
+   */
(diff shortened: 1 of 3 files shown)


padamstx:
I was thinking of the compatibility angle from the standpoint of the generated code, but I can see that if someone was (somehow) calling the HttpClientSingleton explicitly, then they could be broken.
I think the BaseService changes are in fact backwards compatible though.
I can add back the old version of the HttpClientSingleton methods and just code them to call the new methods with the correct arguments.

lpatino10:
Whoops, you're right and it doesn't look like there's actually anything breaking in here 🙃  I must've read those signatures wrong yesterday but you're right that they weren't accessible.

All looks good after that correction!

PR #37, opened 2019-05-25
feat(auth): re-organize authentication function and support ICP4D authentication

padamstx:
Fixes: https://github.ibm.com/watson/developer-experience/issues/6490

Why:
We need to provide a tactical solution for supporting the ICP4D authentication mechanism

What:
This PR includes a re-packaging of the authentication function within the java core
to allow it to be extended with additional authentication implementations, plus the
implementation of the new ICP4D authentication type.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
@@ -25,6 +25,12 @@ import com.ibm.cloud.sdk.core.http.HttpStatus;
 import com.ibm.cloud.sdk.core.http.ResponseConverter;
 import com.ibm.cloud.sdk.core.http.ServiceCall;
 import com.ibm.cloud.sdk.core.http.ServiceCallback;
+import com.ibm.cloud.sdk.core.security.Authenticator;
+import com.ibm.cloud.sdk.core.security.AuthenticatorConfig;
+import com.ibm.cloud.sdk.core.security.AuthenticatorFactory;
+import com.ibm.cloud.sdk.core.security.basicauth.BasicAuthConfig;
+import com.ibm.cloud.sdk.core.security.noauth.NoauthAuthenticator;
+import com.ibm.cloud.sdk.core.security.noauth.NoauthConfig;
 import com.ibm.cloud.sdk.core.service.exception.BadRequestException;
 import com.ibm.cloud.sdk.core.service.exception.ConflictException;
 import com.ibm.cloud.sdk.core.service.exception.ForbiddenException;
@@ -39,10 +45,6 @@ import com.ibm.cloud.sdk.core.service.exception.UnsupportedException;
 import com.ibm.cloud.sdk.core.service.security.IamOptions;
 import com.ibm.cloud.sdk.core.util.CredentialUtils;
 import com.ibm.cloud.sdk.core.util.RequestUtils;
-import com.ibm.cloud.sdk.security.Authenticator;
-import com.ibm.cloud.sdk.security.AuthenticatorConfig;
-import com.ibm.cloud.sdk.security.AuthenticatorFactory;
-import com.ibm.cloud.sdk.security.basicauth.BasicAuthConfig;
 
 import io.reactivex.Single;
 import okhttp3.Call;
@@ -327,7 +329,7 @@ public abstract class BaseService {
    * @param builder the new authentication
    */
   protected void setAuthentication(final Builder builder) {
-    if (skipAuthentication) {
+    if (this.skipAuthentication) {
       return;
     }
 
@@ -423,11 +425,21 @@ public abstract class BaseService {
   protected void setAuthenticator(AuthenticatorConfig authConfig) {
     try {
       this.authenticator = AuthenticatorFactory.getAuthenticator(authConfig);
+      if (authenticator instanceof NoauthAuthenticator) {
+        setSkipAuthentication(true);
+      }
     } catch (Exception e) {
       throw new RuntimeException(e);
     }
   }
 
+  /**
+   * Returns the Authenticator instance currently set on this BaseService instance.
+   * @return the Authenticator set on this BaseService
+   */
+  protected Authenticator getAuthenticator() {
+    return this.authenticator;
+  }
 
   /*
    * (non-Javadoc)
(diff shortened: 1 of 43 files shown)


padamstx:
@mkistler @dpopp07 @lpatino10 This is an in-progress PR that contains my changes (so far) for re-organizanizing the authentication function within the Java core to support ICP4D.   I wanted to get this first bit of work pushed out so you guys could take a look and let me know if you have any feedack so far.   
I still have the following left to do:
- Refactor the IamTokenManager class to put most of it in a base class so it can be shared with the ICP4D token manager implementation (ICP4DAuthenticator class).
- Add tests for all the new function
- Add tests (or modify existing tests) to verify that outgoing requests do in fact have a valid Authorization header attached to them.  I don't think our existing tests do that just yet.

padamstx:
cc: @germanattanasio

padamstx:
Note: there will be some minor tweaks required in the SDK generator to support this new style... specifically the service ctor that is generated to take in the IamOptions argument will need to be changed to take in an AuthenticatorConfig instead.  This *should* be a compatible change wrt the generated code (famous last words).   My goal with the java core changes is to not require changes to existing apps that are already using IAM or Basic Auth.

lpatino10:
We're trying to add this feature this week for Java while Phil's out, so I'll go ahead and take over finishing this PR.

src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
```diff
@@ -397,14 +421,27 @@ public void setDefaultHeaders(final Map<String, String> headers) {
...
   }
 
-  private void clearIamCredentials() {
-    this.tokenManager = null;
+  /**
+   * Initializes a new Authenticator instance based on the input AuthenticatorConfig instance and sets it as
+   * the current authenticator on the BaseService instance.
+   * @param authConfig the AuthenticatorConfig instance containing the authentication configuration
+   */
+  protected void setAuthenticator(AuthenticatorConfig authConfig) {
```
lpatino10: This should be `public` if we're directing users here from something like `setUsernameAndPassword()`

lpatino10:
My thought now to build on this is to make `setAuthenticator()` public and add a new service constructor with the following format in the generator (using Assistant as an example):

```java
public Assistant(String versionDate, AuthenticatorConfig authConfig) {
  // constructor logic
}
```

We could get away with just that constructor now since the other configuration options like username/password and IAM all implement `AuthenticatorConfig`. We could then just deprecate the old constructors that deal with authentication until the next major release.

src/main/java/com/ibm/cloud/sdk/security/AuthenticatorFactory.java
```diff
@@ -0,0 +1,63 @@
...
+  public static Authenticator getAuthenticator(AuthenticatorConfig config) throws IllegalArgumentException {
+
+    // Validate the configuration passed in.
+    config.validate();
+
+    // Create an authenticator from the config.
+    // TODO in the future, we could improve this factory so that it "discovers" Authenticator
+    // implementations (perhaps via the ServiceLoader mechanism), rather than explicitly checking
+    // for specific ones here.
+    if (config instanceof IamOptions) {
```
lpatino10: I think it might make more sense to just compare the `authenticationType` coming from the `AuthenticatorConfig`

src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
padamstx: > We're trying to add this feature this week for Java while Phil's out, so I'll go ahead and take over finishing this PR.

I'm continuing to work on it so should have something by end of week.

padamstx:
> My thought now to build on this is to make `setAuthenticator()` public and add a new service constructor with the following format in the generator (using Assistant as an example):
> 
> ```java
> public Assistant(String versionDate, AuthenticatorConfig authConfig) {
>   // constructor logic
> }
> ```
> 
> We could get away with just that constructor now since the other configuration options like username/password and IAM all implement `AuthenticatorConfig`. We could then just deprecate the old constructors that deal with authentication until the next major release.

The plan is to change the constructor that we already generate (that takes an IamOptions instance as input) and modify it to so the input param is an AuthenticatorConfig instance, which would mean existing apps can continue to pass in an IamOptions object.

src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
padamstx: > This should be `public` if we're directing users here from something like `setUsernameAndPassword()`

Actually my intent was for setAuthenticator() to be called within the generated service class ctor.

lpatino10:
Sounds good @padamstx I'll leave it for you to finish then 👍  It all looks good to me so far.

Just regarding `setAuthenticator()` being called from the generated constructors, do you think we also want some equivalent way to set the auth other than the constructor? Since we currently allow that with U/P and IAM.

padamstx:
> Sounds good @padamstx I'll leave it for you to finish then +1 It all looks good to me so far.
> 
> Just regarding `setAuthenticator()` being called from the generated constructors, do you think we also want some equivalent way to set the auth other than the constructor? Since we currently allow that with U/P and IAM.

Perhaps... let's see where we end up when things are mostly finished and we can decide whether additional methods need to be exposed to apps.   We could expose something like BaseService.setAuthenticatorConfig(AuthenticatorConfig) that would allow the app to replace whatever might be in place for authentication within the BaseService instance with a new config, etc.

germanattanasio:
Can we call it `Authentication` rather than `AuthenticationConfig`?

lpatino10:
I'm a fan of `AuthenticationStrategy` actually 👀

padamstx:
> I'm a fan of `AuthenticationStrategy` actually eyes

If y'all like `AuthenticationStrategy` that much better than `AuthenticatorConfig`, go for it :)
I guess it does sound a little fancier I guess (although the "Config" part is descriptive of precisely what is stored in the object).
lol

lpatino10:
Just finished all the tests, so I think this is ready for a serious look. I didn't end up renaming those config classes but we can if there's a strong opinion around it.

src/test/java/com/ibm/cloud/sdk/core/test/service/AuthenticationTest.java
```diff
@@ -19,20 +31,69 @@
...
   }
 
   @Test
-  public void authenticateWithApiKeyAsUsername() {
+  public void authenticateIam() {
+    IamOptions options = new IamOptions.Builder()
+        .apiKey(APIKEY)
+        .build();
+    TestService service = new TestService(options);
+    service.setUsernameAndPassword(APIKEY_USERNAME, APIKEY);
```
lpatino10: I don't think this line should be in this test

lpatino10:
@padamstx just did the live test and all the ICP4D auth logic works as expected 🙌

PR #39, opened 2019-06-07
Added HTTP status code to response

ricellis:
The `Response` class is usually only generated for a successful response (i.e. `200` <= status code < `400`), but there may still be useful information encapsulated in the successful status code. For example, when creating resources `201` and `202` convey different information about what has happened on the server (`Created` vs `Accepted`).

This PR adds a `com.ibm.cloud.sdk.core.http.Response#getStatusCode` method so that additional information of the status code can be accessed by the caller.

It adds a single test `com.ibm.cloud.sdk.core.test.service.ResponseTest#testResponseCode` to demonstrate that the code is set and retrievable.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/Response.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/Response.java
@@ -22,11 +22,13 @@ public class Response<T> {
   private T result;
   private Headers headers;
   private int statusCode;
+  private String statusMessage;
 
   public Response(T result, okhttp3.Response httpResponse) {
     this.result = result;
     this.headers = new Headers(httpResponse.headers());
     this.statusCode = httpResponse.code();
+    this.statusMessage = httpResponse.message();
   }
 
   public T getResult() {
@@ -39,5 +41,9 @@ public class Response<T> {
 
   public int getStatusCode() {
     return this.statusCode;
+  }
+
+  public String getStatusMessage() {
+    return this.statusMessage;
   }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/ibm/cloud/sdk/core/http/Response.java
```diff
@@ -21,10 +21,12 @@
 
   private T result;
   private Headers headers;
+  private int statusCode;
 
   public Response(T result, okhttp3.Response httpResponse) {
     this.result = result;
     this.headers = new Headers(httpResponse.headers());
+    this.statusCode = httpResponse.code();
```
padamstx: While we're at it, would it also make sense to save off the httpResponse.message() value (status message)?

ricellis: The [spec](https://tools.ietf.org/html/rfc7230#section-3.1.2) calls that part a "reason-phrase"; what is your preferred naming here:
* `getReasonPhrase()`
* `getMessage()`
* `getStatusMessage()`
* something else?

padamstx: My initial response was going to be "statusMessage" because it will be somewhat in sync with "statusCode" and they are certainly related, but alas the Http spec braintrust has foiled our naming attempts...  Since the spec calls it a reason phrase, I guess we should do that as well:  `getReasonPhrase()`

padamstx: Talked with Rich on slack a bit and we convinced ourselves that getStatusMesage() would be more intuitive to users since it goes along with the status code.

ricellis: Added `getStatusMessage()` in c7c0b6a.

PR #41, opened 2019-06-27
feat(serialization): Allow dynamic models to have non-public default ctors

padamstx:
This PR will allow dynamic models (subclasses of DynamicModel<T>) to define a protected (or private) default ctor instead of a public one.   We need this to accommodate the code pattern used by our new oneOf/anyOf processing whereby the parent schema (the schema containing the oneOf/anyOf list) will define a protected default ctor to prevent users from instantiating it (they would have to use a base class instead).   For this reason, we need the DynamicModelTypeAdapterFactory class to recognize a non-public default ctor when checking to see if it should handle a specific class.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/DynamicModelTypeAdapterFactory.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/DynamicModelTypeAdapterFactory.java
@@ -106,7 +106,7 @@ public class DynamicModelTypeAdapterFactory implements TypeAdapterFactory {
     Constructor<?> ctor = getDefaultCtor(rawType);
     if (ctor == null) {
       LOGGER.warning("Instance of class " + rawType.getName() + " is a subclass of DynamicModel, but it doesn't "
-        + "have a public default constructor.  This instance will be ignored by " + this.getClass().getSimpleName());
+        + "define a default constructor.  This instance will be ignored by " + this.getClass().getSimpleName());
       return null;
     }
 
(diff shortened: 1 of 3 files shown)


lpatino10:
Hey Phil, I get the rationale of needing a protected constructor for the `oneOf`/`anyOf` support from your demo yesterday, but I'm not 100% clear on the impact to the generated Watson Java SDK as it stands now. This won't **remove** any generated constructors, just allow for new ones, right?

Just want to make sure it's not a breaking change or anything, because otherwise it looks good to me.

padamstx:
This change in the java core simply makes the DynamicModelTypeAdapterFactory work correctly if the model class in question has a non-public default ctor, where previously it would only recognize models with a public default ctor.   So this, in and of itself, doesn't change any generated code or introduce any sort of breaking changes.  Instead, it will be needed for situations where we generate a base class for a schema that uses oneOf (and also uses additionalProperties, making it a dynamic model) because in that narrow case we'll make the default ctor protected (instead of public) to prevent users from instantiating the base class directly.

tl;dr... no this won't introduce any breaking changes :)

germanattanasio:
Will this affect users trying to set values in the assistant context and send it back as a new message? I've seen users having issues using the setters in subsequent requests to `/message`

padamstx:
> Will this affect users trying to set values in the assistant context and send it back as a new message? I've seen users having issues using the setters in subsequent requests to `/message`

No :)

PR #47, opened 2019-07-12
feat(serialization): added gson type adapter for byte[]

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/908

This PR introduces a Gson Type Adapter for "byte []" values.
This will ensure that any model class fields of type "byte[]" are serialized and deserialized properly.
In the case of "byte[]" this means that the sequence of bytes is transformed into a string of base64-encoded characters.   On the deserialization side, we would transform the string of base64-encoded characters found in the JSON instance by decoding them into their original byte values and store in the resulting byte[].

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/GsonSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/GsonSingleton.java
@@ -52,6 +52,7 @@ public final class GsonSingleton {
     // Date serializer and deserializer
     builder.registerTypeAdapter(Date.class, new DateDeserializer());
     builder.registerTypeAdapter(Date.class, new DateSerializer());
+    builder.registerTypeAdapter(byte[].class, new ByteArrayTypeAdapter());
 
     // Type adapter factory for DynamicModel subclasses.
     builder.registerTypeAdapterFactory(new DynamicModelTypeAdapterFactory());
--- a/pom.xml
+++ b/pom.xml
@@ -46,6 +46,7 @@
         <logback-version>1.2.3</logback-version>
         <guava-version>27.1-android</guava-version>
         <gson-version>2.8.5</gson-version>
+        <commons-codec-version>1.12</commons-codec-version>
         <commons-io-version>2.6</commons-io-version>
         <commons-lang3-version>3.8.1</commons-lang3-version>
         <rx-version>2.2.7</rx-version>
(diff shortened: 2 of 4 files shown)


padamstx:
cc: @dpopp07

padamstx:
Uh oh, looks like the Java7 build failed due to "Base64".   I'll need to find an equivalent that works for java 7 *sigh* :)

padamstx:
Pushed an amended commit which includes a change to use the Base64 class from apache commons codec package instead of from "java.util".

src/main/java/com/ibm/cloud/sdk/core/util/GsonSingleton.java
```diff
@@ -52,6 +52,7 @@ private static void registerTypeAdapters(GsonBuilder builder) {
     // Date serializer and deserializer
     builder.registerTypeAdapter(Date.class, new DateDeserializer());
     builder.registerTypeAdapter(Date.class, new DateSerializer());
+    builder.registerTypeAdapter(byte[].class, new ByteArrayTypeAdapter());
```
ricellis: I don't recall; is another type (e.g. `File`) being used for the alternative cases where `format: binary`?

If not then I don't think this type adapter should be registered "globally" because there may be some `byte[]` that need to be mapped with `type: string, format: binary` as opposed to `format: base64` (which was why I'd originally suggested annotating the fields). I can't find anything about what JSON string is actually expected when using `format: binary` other than "any sequence of octets", which I guess could just be the corresponding UTF-8 characters.

I agree that registering an adapter globally is a nicer solution, but we just need to take care that we don't break the other format by doing it.

mkistler: It looks to me like `type: string, format: byte`, which the openAPI generator calls `ByteArray`, is the only type mapped to `byte[]`, so it seems safe to me to register this type adapter globally.

padamstx: Correct, here is the complete type mapping for the Java generator:
```
        typeMapping.put("BigDecimal", "BigDecimal");
        typeMapping.put("ByteArray", "byte[]");
        typeMapping.put("file", "InputStream");
        typeMapping.put("binary", "InputStream");
        typeMapping.put("boolean", "Boolean");
        typeMapping.put("date", "Date");
        typeMapping.put("DateTime", "Date");
        typeMapping.put("number", "Double");
        typeMapping.put("float", "Float");
        typeMapping.put("double", "Double");
        typeMapping.put("long", "Long");
        typeMapping.put("integer", "Long"); // Use only 64-bit integers
        typeMapping.put("map", "Map");
        typeMapping.put("array", "List");
        typeMapping.put("UUID", "String");
        typeMapping.put("string", "String");
        typeMapping.put("object", "Object");
        typeMapping.put("any", "Object");
```
the Java "byte[]" type is only used for the openapi ByteArray type (which refers to a schema with "type string/format byte".

PR #49, opened 2019-07-23
feat(serialization): Avoid ser/deser roundtripping asymmetry for Object map value type

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/951

This PR introduces changes to avoid ser/deser asymmetry problems when dealing with 
DynamicModels with "Object" as the map value type.
Summary of changes:
1) Introduced new "MapValueObjectTypeAdapter" which will be used specifically for map values when
the map value type is Object.   This is an adaptation of Gson's ObjectTypeAdapter, with the main change being that a JSON number will be deserialized into an instance of Gson's LazilyParseNumber instead of java.lang.Double.
2) DynamicModelTypeAdapterFactory was changed to use MapValueObjectTypeAdapter when the map value type is Object.
3) The relevant testcase was modified to NOT compensate for the unnecessary decimal points when serializing a previously deserialized object, since the decimal points will no longer be included in the JSON string.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/DynamicModelTypeAdapterFactory.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/DynamicModelTypeAdapterFactory.java
@@ -296,11 +296,13 @@ public class DynamicModelTypeAdapterFactory implements TypeAdapterFactory {
     private Constructor<?> ctor;
     private Map<String, BoundField> boundFields;
     private Gson gson;
+    private TypeAdapter<?> mapValueObjectTypeAdapter;
 
     Adapter(Gson gson, Constructor<?> ctor, Map<String, BoundField> boundFields) {
       this.gson = gson;
       this.ctor = ctor;
       this.boundFields = boundFields;
+      this.mapValueObjectTypeAdapter = new MapValueObjectTypeAdapter(gson);
     }
 
     /*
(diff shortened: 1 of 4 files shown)


padamstx:
cc: @ricellis

src/test/java/com/ibm/cloud/sdk/core/test/model/DynamicModelSerializationTest.java
```diff
@@ -261,7 +261,7 @@ public void testModelAPString() {
   @Test
   public void testModelAPFooNullTypeToken() {
     ModelAPFooNullTypeToken model = createModelAPFooNullTypeToken();
-    testSerDeser(model, ModelAPFooNullTypeToken.class, true);
+    testSerDeser(model, ModelAPFooNullTypeToken.class, false);
```
mkistler: This change eliminates the need to ever "trim decimals" in `testSerDeser`, so lets just clean that up as part of this PR.  Meaning ... eliminate that parameter and the associated logic from `testSerDeser` and all the call sites.

src/main/java/com/ibm/cloud/sdk/core/util/MapValueObjectTypeAdapter.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright (C) 2011 Google Inc.
```
lpatino10: Should we still have a Google copyright?

src/test/java/com/ibm/cloud/sdk/core/test/model/DynamicModelSerializationTest.java
lpatino10: ☝️ agreed

src/main/java/com/ibm/cloud/sdk/core/util/DynamicModelTypeAdapterFactory.java
```diff
@@ -385,7 +403,11 @@ public T read(JsonReader in) throws IOException {
           } else {
             // Otherwise, it must be an additional property so it belongs in the map.
             String key = name;
-            Object value = mapValueTypeAdapter.read(in);
+            Object value;
+
+            value = mapValueTypeAdapter.read(in);
```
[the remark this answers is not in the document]
padamstx: I originally had some code between lines 406 and 408, but removed them in preference to a better solution but neglected to combine those lines again :)

src/main/java/com/ibm/cloud/sdk/core/util/MapValueObjectTypeAdapter.java
padamstx: You're right, I should add our own at the top.   But perhaps we should also include the original?   I essentially copied a class from GSON and then made some minor changes to it.

padamstx:
Pushed an amended commit that should address all the review comments...

PR #50, opened 2019-08-01
fix: do not use 'serializeNulls' when initializing Gson

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/960

This PR removes the call to `serializeNulls()` when creating a `Gson` instance while serializing JSON content inside `RequestBuilder.bodyContent(String, Object, Object, InputStream)`.
This will ensure that we're serializing JSON body content in the same way no matter whether the operation accepts only JSON content, or JSON content in addition to other types as well.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/RequestBuilder.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/RequestBuilder.java
@@ -334,7 +334,7 @@ public class RequestBuilder {
   public RequestBuilder bodyContent(String contentType, Object jsonContent, Object jsonPatchContent,
     InputStream nonJsonContent) {
     if (contentType != null) {
-      Gson requestGson = GsonSingleton.getGson().newBuilder().serializeNulls().create();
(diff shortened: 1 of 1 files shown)


lpatino10:
I think we might need to do something elsewhere to preserve this functionality. We originally made the decision to include `null` values in responses based on this thread: https://github.com/watson-developer-cloud/java-sdk/issues/959

Could you elaborate more on why we might need to change the current implementation?

padamstx:
This PR fixes issue https://github.ibm.com/arf/planning-sdk-squad/issues/960 reported by the Cloudant team.   The gist of the problem is that there is a discrepancy between the serialization behavior for an operation that has only application/json body content vs an operation that has application/json AND other body content types.
Specifically, the `RequestBuilder.bodyContent(String, Object, Object, InputStream)` method is calling `GsonBuilder.serializeNulls()`, whereas the alternate serialization path used for an operation that has only application/json body content is not.   These two different serialization paths should be consistent, IMO.
This doesn't affect the deserialization of responses at all.

PR #55, opened 2019-08-28
fix: perform correct transformation on serviceName; use case-insensitive auth type comparisons

padamstx:
This PR does the following:
1. When using the service name to retrieve external configuration properties (credential file or environment variables) any "-" (dash) characters are replaced with an "_" (underscore) character (in addition to folding the service name to upper case).
2. When using the <servicename>_AUTH_TYPE config property to determine which type authenticator should be created, we'll now use a case-insensitive comparison to be a little more lenient.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/Headers.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/Headers.java
@@ -1,3 +1,16 @@
+/**
+ * (C) Copyright IBM Corp. 2015, 2019.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
 package com.ibm.cloud.sdk.core.http;
 
 import java.util.List;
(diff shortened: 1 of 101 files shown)


lpatino10:
Also small thing, but I realized that in the method Javadoc on line 141, we still reference JNDI. We should remove that.

padamstx:
> For the name substitution (hyphen -> underscore), should we be doing that in all three places? And by that I mean when looking in the file, environment, and VCAP. It looks like in the changes now, only the first two are covered.

This is by design as we want to use the x-vcap-service-name api def annotation value as-is when searching for services in VCAP_SERVICES.   For cred file property names and env var names, we need to do the transformation.  I actually started to change the code exactly in the way you are suggesting but realized that the VCAP stuff is the odd man out.

> 
> Also, if it is the case that they should all get the same treatment, I wonder if we should consider making a lot of the individual methods in `CredentialUtils` private. 

Yes, I think we can expose only getServiceProperties from CredentialUtils and make the rest private.

lpatino10:
If that's the case, then we're relying on the VCAP service name being passed in in the first place to get the credentials, but at the moment it's not, so something else will need to change I think.

padamstx:
@lpatino10 Ok, I think this is ready for a review now :)

padamstx:
> I was told we were going to be discussing the service name we pass in from the generator in today's meeting, so we can get that cleared up. As is, I think there will be a discrepancy for some services in VCAP (e.g. visual_recognition vs. watson_vision_combined).

HAHA, yes, I did a survey of the various cores yesterday and noticed that some have the watson hack and some do not.   I would vote for removing that "visual_recognition" --> "watson_vision_combined" hack everywhere and just be consistent with the service name key used for loading properties... i.e. change the VCAP_SERVICES function to use "visual_recognition" or change the credential file function to use "watson_vision_combined" (and change the x-vcap-service-name value in the API def to be "watson_vision_combined" as well).

PR #59, opened 2019-09-16
Change order of files checked for authentication

lpatino10:
Related to https://github.ibm.com/Watson/developer-experience/issues/7520

This PR changes the order in which we search for external credential files, moving the search for a file in the current working directory up just after the user-defined path.


lpatino10:
@padamstx I didn't make that slight logic change Taj did in [this commit](https://github.com/IBM/dotnet-sdk-core/pull/28/commits/947bcf2d57e2a096168e8f8203365e05732a22f4#diff-1b3dca31b3bc6842307ad980e59f4b03) because our code was set up a bit differently. He was doing validation of the file in that first method that got simplified, while I'm making sure it's a valid file in the `getFirstExistingFileContents` method.

The change didn't really simplify things in this case, just move code around, so I decided to leave it.

PR #60, opened 2019-09-18
Give more information about SSL handshake exception for ICP/CP4D

lpatino10:
Related to https://github.ibm.com/Watson/developer-experience/issues/7541

This PR adds custom warnings for cases when we run into an `SSLHandshakeException`. This could happen for users calling service endpoint on ICP or CP4D because they haven't set up self-signed certificates.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
@@ -63,8 +63,8 @@ public abstract class BaseService {
 
   private static final Logger LOG = Logger.getLogger(BaseService.class.getName());
 
-  private static final String ERROR_MSG_NO_AUTHENTICATOR = "Authentication information was not properly configured.";
-  private static final String ERROR_MSG_SSL = "If you're trying to call a service on ICP or Cloud Pak for Data, you "
+  private static final String ERRORMSG_NO_AUTHENTICATOR = "Authentication information was not properly configured.";
+  private static final String ERRORMSG_SSL = "If you're trying to call a service on ICP or Cloud Pak for Data, you "
       + "may not have a valid SSL certificate. If you need to access the service without setting that up, try using "
       + "the disableSsl option in your authentication configuration and/or the disableSslVerification option in the "
       + "HttpConfigOptions.";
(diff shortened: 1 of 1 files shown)


src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
```diff
@@ -435,10 +447,17 @@ public void onResponse(Call call, Response response) {
...
-        public com.ibm.cloud.sdk.core.http.Response<T> call() throws Exception {
-          Response response = call.execute();
-          T responseModel = processServiceCall(converter, response);
-          return new com.ibm.cloud.sdk.core.http.Response<>(responseModel, response);
+        public com.ibm.cloud.sdk.core.http.Response<T> call() {
+          try {
+            Response response = call.execute();
+            T responseModel = processServiceCall(converter, response);
+            return new com.ibm.cloud.sdk.core.http.Response<>(responseModel, response);
+          } catch (IOException e) {
```
padamstx: Do we truly want to catch and re-throw only IOExceptions as a RuntimeException?   Since you removed the throws class from the function signature, shouldn't we catch all exceptions (i.e. Throwable) and then re-throw as a RuntimeException?

src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
```diff
@@ -61,7 +63,11 @@
 
   private static final Logger LOG = Logger.getLogger(BaseService.class.getName());
 
-  private static final String ERRORMSG_NO_AUTHENTICATOR = "Authentication information was not properly configured.";
+  private static final String ERROR_MSG_NO_AUTHENTICATOR = "Authentication information was not properly configured.";
```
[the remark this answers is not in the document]
lpatino10: That's fair. Just changed those back 👍

src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
lpatino10: `enqueue` needs an `IOException` in the overridden signature for `onFailure`, but I went ahead and changed it to throw all `Exception`s for `execute` and `reactiveRequest`.

lpatino10: Actually, I think we should stick with just throwing `IOException`s. The `processServiceCall` method will handle returning the more specific exceptions like 404s and 500s. With the change I just made, tests that are looking for more specific errors are breaking.

padamstx: Ok, do what you need to do.   I added my original comment because I noticed the potential change in behavior is all.

lpatino10:
I agree! Consistency is best

PR #61, opened 2019-09-23
New FileWithMetadata class to support arrays of files in form data

mkistler:
This PR adds the "FileWithMetadata" class to the "models" package of the core for use as the item value in an array of files parameter.


mkistler:
@padamstx Good thought, but I don't think there is much to simplify in the generated code. We will loop for the array and for each item we do:
```
    okhttp3.RequestBody itemBody = RequestUtils.inputStreamBody(item.data, item.contentType);
    multipartBuilder.addFormDataPart("{{baseName}}", item.filename, itemBody);
```
I suppose we could add a method `RequestUtils.fileWithMetadataBody()`, but I think the benefit there is marginal.

mkistler:
Tests added.  In the process of writing the tests I found a design issue with the main class -- it was missing setters for the `data` property.

PR #66, opened 2019-10-28
feat(BaseService): add new method ConfigureService() to BaseService class

jorge-ibm:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/1120

Changes:
- moved the external configuration code outside of the constructor into it's own method.
- for compatibility, the constructor calls the new method until we move to the service factory's complete implementation.
- created a new test case for the configureservice method since PowerMock is dependent on JUnit, and the BaseServiceTest class was written using TestNG. PowerMock was needed to mock environment variables

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
@@ -101,15 +101,23 @@ public abstract class BaseService {
     }
     this.authenticator = authenticator;
 
+    // Configure a default client instance.
+    this.client = configureHttpClient();
+
+    // temp: set any external configuration from the constructor
+    configureService(name);
+  }
+
+  protected void configureService(String serviceName) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
```diff
@@ -101,16 +101,24 @@ public BaseService(final String name, Authenticator authenticator) {
     }
     this.authenticator = authenticator;
 
+    // Configure a default client instance.
+    this.client = configureHttpClient();
```
jorge-ibm: @padamstx - just to be sure, the client initiating should not be dependent on external config correct?

padamstx: Correct, "configureHttpClient()" does not involve external config stuff.

PR #67, opened 2019-10-29
feat(serialization): add gson type adapter factory to handle discriminator metadata

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/970

This PR provides the required support within the Java core to support discriminators in composed schemas.   Specifically, there's a new type adapter factory that will be registered with Gson.  The factory will create TypeAdapter instances to handle classes that contain discriminator metadata.

There will be a corresponding PR for the Java generator that depends on this PR, but this PR stands on its own (no dependencies).


padamstx:
1. Yes, the trace logging could be useful in debugging problems in the field.  I didn't actually "need" to check in the logging.properties file but it won't be something that's delivered to the user anyhow so shouldn't create problems.   The "fine" and "finest" logging will only occur if explicitly enabled by the user, which we might ask someone to do if they encounter problems with the deserialization behavior.  My prior life as a backend developer is showing a little on this one :)

2. Not sure I understand the question.   The generation of the new static fields will occur ONLY if the parent schema defines the "discriminator" object.   By "custom annotation" did you mean in the API definition or a Java annotation???

3. The classes were all copied from the generator unit test output resulting from src/test/resources/discriminator-test.yaml, and that update is in my generator PR.

PR #69, opened 2019-11-06
feat: enhancing vcap parsing & removing configureService reference from BaseService ctor

jorge-ibm:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1098

The vcap changes were based on @padamstx initial implementation in the service-factory branch. Thanks Phil!

Changes:
- removed the call to the ``configureService()`` method out of the constructor.
- updated the unit tests that were dependent on ``configureService()`` being called when instantiating the BaseService.
- updated how external configuration is loaded from a vcap file based on this [design](https://github.ibm.com/CloudEngineering/openapi-sdkgen/wiki/Design-Of-Service-Config-Changes]). Specifically, the parsing now checks for a matching service name within an entry, if it doesn't find any then it will return the first entry with a matching outer key to the service name.
- other general cleanup of unused code

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
@@ -103,12 +103,9 @@ public abstract class BaseService {
 
     // Configure a default client instance.
     this.client = configureHttpClient();
-
-    // temp: set any external configuration from the constructor
-    configureService(name);
   }
 
-  protected void configureService(String serviceName) {
+  public void configureService(String serviceName) {
     if (serviceName == null || serviceName.isEmpty()) {
       throw new IllegalArgumentException("Error configuring service. Service name is required.");
     }
--- a/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
@@ -17,25 +17,24 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
-import com.google.gson.JsonSyntaxException;
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
 import com.ibm.cloud.sdk.core.security.Authenticator;
 import com.ibm.cloud.sdk.core.service.BaseService;
+
 
 /**
  * CredentialUtils retrieves service credentials from the environment.
@@ -46,18 +45,7 @@ public final class CredentialUtils {
   public static final String PLAN_STANDARD = "standard";
 
   private static final String DEFAULT_CREDENTIAL_FILE_NAME = "ibm-credentials.env";
-
   private static final String VCAP_SERVICES = "VCAP_SERVICES";
-
-  private static final String CREDENTIALS = "credentials";
-  private static final String PLAN = "plan";
-  private static final String USERNAME = "username";
-  private static final String PASSWORD = "password";
-  private static final String URL = "url";
-  private static final String IAM_APIKEY = "iam_apikey";
-  // this value was used previously for IAM API keys as well
-  private static final String APIKEY = "apikey";
-  private static final String IAM_URL = "iam_url";
 
   private CredentialUtils() {
     // This is a utility class - no instantiation allowed.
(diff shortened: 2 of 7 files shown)


src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
```diff
@@ -108,84 +96,96 @@ public static boolean hasBadStartOrEndChar(String credentialValue) {
...
+   * @return a VcapService instance if found, or null
    */
-  public static String getVcapValue(String serviceName, String key, String plan) {
-    if ((serviceName == null) || serviceName.isEmpty()) {
-      return null;
-    }
-
-    final JsonObject services = getVcapServices();
-    if (services == null) {
+  protected static VcapService getVcapServiceEntry(String serviceName) {
```
jorge-ibm: @padamstx @mkistler - I've made this method protected vs public. Though I don't think users will actually be calling this function themselves. Please let me know otherwise

src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
```diff
@@ -108,84 +96,96 @@ public static boolean hasBadStartOrEndChar(String credentialValue) {
...
+        for (VcapService service : serviceList) {
+          if (serviceName.equals(service.name)) {
+            return service;
+          }
+        }
+      }
+      // Second, try to find a service list with the specified key.
+      List<VcapService> services = vcapObj.get(serviceName);
+      if (services != null && services.size() > 0) {
+        VcapService service = (services.get(0) != null) ? services.get(0) : null;
```
padamstx:
```suggestion
        return services.get(0);
```
this will return either null or a non-null value, which is what we want in both cases

src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
```diff
@@ -108,84 +96,96 @@ public static boolean hasBadStartOrEndChar(String credentialValue) {
...
+        return service;
```
padamstx: with my suggestion above, just remove this line

src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
jorge-ibm: Good point. I went ahead and updated that

jorge-ibm:
@padamstx - I've added some more unit tests that test an unconfigured and a configured service as you requested.

src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
```diff
@@ -103,9 +103,6 @@ public BaseService(final String name, Authenticator authenticator) {
 
     // Configure a default client instance.
     this.client = configureHttpClient();
-
-    // temp: set any external configuration from the constructor
-    configureService(name);
```
mkistler: What are the implications of removing this call here?  Won't this result in different, possibly incompatible results for existing users?

jorge-ibm: Well this idea is if the user specifies ``includeExternalConfig`` to be true, then the constructor in the generated code will make this call to ``configureService`` and not the constructor in the BaseService. However, if a user uses a version of the generator that doesn't have that change, and uses this new version of the core, then yes it would be a breaking change. I think the idea was to be specific about that in the compatibility chart.

jorge-ibm: @mkistler - though I'm still trying to grasp on how you guys do version releases. I think before, if a user didn't specify ``includeExternalConfig`` to be true, the code that looks for external config was still being ran, but it wouldn't actually configure the service. Is that correct?

padamstx: Yes, this will be a change that will affect users.   This new version of the java core will need to be versioned as a breaking change, then when the new version of the generator is tagged (2.2.0?) we'll need to update the compatibility chart accordingly.
For Watson services (i.e. includeExternalConfig is enabled), it won't constitute a breaking change because the new version of the generator will generate their ctors such that they call configureService().   
Only non-Watson services will be affected, but only if applications try to use the new java core with old generated code.  Clear as mud, right?

We also have another related issue that would be good for all three of us to talk about when our schedules line up.

padamstx:
@jorge-ibm before you merge this PR in, could you please make the `BaseService.configureService()` method public instead of protected?   After our discussion yesterday afternoon, I think we're going to want to give the SDK user the ability to call it directly.

lpatino10:
If this is going to be a major release, could we release a new minor version of the core before bringing this change in? There are some things that have been added since the 7.0.0 release that I'd like to bring into the Watson SDK.

padamstx:
> If this is going to be a major release, could we release a new minor version of the core before bringing this change in? There are some things that have been added since the 7.0.0 release that I'd like to bring into the Watson SDK.

Yes, we could do that, but the next version of the generator (to be used for the watson sdk delivery next week) will need to be paired with this next major release of the java core.   So assuming that you use the to-be-delivered-soon new release of the generator to produce your next version of the watson java sdk, you'd need to use the new major version (8.0.0) of the java core for that.

lpatino10:
I was going to make a patch release around the dependency cleanup PR I submitted and a bug fix in the SDK I'm going to be looking into today, so I'd rather get that out when it's ready as opposed to waiting for the next minor release.

Also just for clarity: I'm slightly confused by [this comment of yours](https://github.com/IBM/java-sdk-core/pull/69#discussion_r343806158), but if I'm understanding correctly, it **won't** cause any issues for users of the Watson SDK? Trying to make sure that upgrading to the major version of the core for that release isn't problematic.

jorge-ibm:
> @jorge-ibm before you merge this PR in, could you please make the `BaseService.configureService()` method public instead of protected? After our discussion yesterday afternoon, I think we're going to want to give the SDK user the ability to call it directly.

sure. This makes sense. I've gone ahead and made that change

padamstx:
> I was going to make a patch release around the dependency cleanup PR I submitted and a bug fix in the SDK I'm going to be looking into today, so I'd rather get that out when it's ready as opposed to waiting for the next minor release.
> 
> Also just for clarity: I'm slightly confused by [this comment of yours](https://github.com/IBM/java-sdk-core/pull/69#discussion_r343806158), but if I'm understanding correctly, it **won't** cause any issues for users of the Watson SDK? Trying to make sure that upgrading to the major version of the core for that release isn't problematic.

Ok, yeah I can see why you're confused :)
This PR will in fact constitute a "breaking change" of sorts for an SDK user if they just simply updated their dependency to point to this new version while still using the existing generated code.  The reason for that is that we're removing the call to configureService() from the BaseService ctor.   That means the BaseService ctor will no longer trigger the loading and use of external config for service-level properties (URL, disable_ssl, etc.).   However, Jorge will be making a corresponding change the Java generator such that the generated service ctor will invoke configureService() instead of relying on the call to "super()" to do that.   This is part of our implementation of the service factory feature.  The design is [here](https://github.ibm.com/CloudEngineering/openapi-sdkgen/wiki/Design-Of-Service-Config-Changes).
So, we don't want a user of the current Java SDK to unilaterally upgrade to this new version of the java core unless that is also coupled with new generated code from the about-to-be-delivered Java generator.   When we deliver the new generator release, we'll update the compatibility matrix to reflect the new major version of the java core.

Due to what you need, I'll go ahead and tag a new minor version (or patch... I'll look at the new commits there to determine which is needed), then update the compatibility chart for the current (2.1.0) generator to reflect this new version.

lpatino10:
Okay, that all makes sense now @padamstx. Thanks for the clarification!

PR #73, opened 2019-11-13
Avoid making excess API calls for auth token exchange

lpatino10:
While investigating https://github.com/watson-developer-cloud/java-sdk/issues/1094, I ran into an issue where if I was trying to make a bunch of API calls on different threads with an IAM-authenticated service, I would run into problems in the token exchange. Each thread would be trying to hit the token exchange service and I would run into rate limitation errors. Below is a screenshot of this is action:

Screen Shot 2019-11-13 at 3 27 27 PM

This PR takes an approach discussed in [this issue](https://github.ibm.com/arf/planning-sdk-squad/issues/1091) and implemented in [this Swift SDK core PR](https://github.com/IBM/swift-sdk-core/pull/9).

The main idea is that if a token is requested when it's in our 20% buffer zone, we can return the stored token and perform a new token request in the background. Methods around requesting and refreshing tokens have been synchronized so that duplicate requests are avoided.


src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -211,6 +215,15 @@ public void authenticate(Builder builder) {
   public String getToken() {
     String token;
 
+    // Wait for token to get stored.
+    while (this.isTokenRequestInProgress) {
```
germanattanasio: Why don't you make `getToken()` synchronized?  See: https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -211,6 +215,15 @@ public void authenticate(Builder builder) {
   public String getToken() {
```
germanattanasio:
```suggestion
  public synchronized String getToken() {
```

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
lpatino10: That makes a lot of sense 🤦‍♀

padamstx:
Currently the java core is at v 8.0.0, and any SDK that uses that version would also need to make sure they use the latest version of the Java generator (we'll tag the new generator release soon).
For this reason, if this fix is needed in the pre-8.0.0 version of the core, we should probably create a new release branch from the 7.1.0 tag value, then cherry-pick this commit to the new branch, then tag a 7.1.1 version from that branch.
If we need this, let me know and I can do the git work to make that happen.

lpatino10:
I do think it's a nice-to-have if you don't mind doing that @padamstx

lpatino10:
I'm all for consistency, but looking at that PR, I can't really tell how that solution is functionally better than adding the `synchronized` keyword, and I don't think we should introduce extra complication if we don't have to. As it stands, this change will always return the current token if it's valid and limit calls to the token service to 1 to avoid any rate limits.

What is the benefit of advancing the refresh time if we just pause threads and don't make the extra API call in the first place?

ricellis:
The reason that solution is functionally better than the `synchronized` keyword is that it doesn't block all requests during the renewal cycle while there is still an _unexpired_ token. `synchronized` is fine for the first exchange when you don't want any requests to happen at all until a token is initially acquired (or in the case where a token is already expired), but imposes a bottleneck during a pre-emptive renewal.

> this change will always return the current token if it's valid

Yes, but valid in this case means < 80% of expiry time _not_ unexpired. Hence I can have an unexpired token that can still successfully be used to make requests, but `synchronized` will block all of those requests while it goes to fetch a new token (that isn't actually needed for another 20% of the lifetime, typically 12 minutes).

lpatino10:
Thanks for the clarification @ricellis. That does make sense.

I'll need to look at the Swift PR a bit more to understand the implementation but I can go ahead and do something similar for this PR.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -211,9 +239,23 @@ public void authenticate(Builder builder) {
...
     String token;
 
-    if (this.tokenData == null || !this.tokenData.isTokenValid()) {
-      // request new token
-      this.tokenData = requestToken();
+    if (this.tokenData == null) {
+      this.tokenData = synchronizedRequestToken();
+    } else if (!this.tokenData.isTokenValid()) {
```
mkistler:
```suggestion
    } else if (!this.needsRefresh()) {
```

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -211,9 +239,23 @@ public void authenticate(Builder builder) {
...
+
+      // Token is "invalid" but it's not expired, meaning it's in our buffer zone. We can still return the stored
+      // token while we go ahead and get a new token in the background.
+      if (!this.tokenData.isTokenExpired()) {
```
mkistler: I don't think you need this check here.  If the token needs to be refreshed, we always want to refresh.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -202,6 +202,34 @@ public void authenticate(Builder builder) {
...
+   * Requests a new token. The refresh time for the token is also updated to some point in the future to prevent
+   * other threads from also calling this method while the new token is set.
+   */
+  private synchronized void refreshToken() {
+    // Another thread may have already updated this time. In that case, we can just return.
+    if (this.tokenData.isTokenValid()) {
+      return;
+    }
+
+    this.tokenData.advanceRefreshTime();
```
mkistler: Rather than advancing the refresh time here, I think you should do it within `needsRefresh` whenever it responds `true`.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
lpatino10: Yeah looking at it again I think you're right. I initially had the check there to make sure that we don't return the stored access token if it's actually completely expired, but I don't think the refresh logic needs to be included in that.

lpatino10:
@mkistler thanks for the review! I agreed with your comments about cleaning things up. I just pushed a new commit addressing those comments when you get the chance to take a look again.

lpatino10:
@ricellis I can't tag you as a reviewer for some reason, but feel free to chime in as well since you were involved in the initial discussions.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -211,12 +236,26 @@ public void authenticate(Builder builder) {
...
-      this.tokenData = requestToken();
+    if (this.tokenData == null) {
+      this.tokenData = synchronizedRequestToken();
+    } else if (this.tokenData.needsRefresh()) {
+
+      // Kick off background task to refresh token.
+      Thread updateTokenCall = new Thread(new Runnable() {
+        @Override
+        public void run() {
+          refreshToken();
```
mkistler: I think you should just call `requestToken` here.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -202,6 +202,31 @@ public void authenticate(Builder builder) {
...
+   * Requests a new token after performing one last check that the token still needs refreshing, in case another thread
+   * has since refreshed it.
+   */
+  private synchronized void refreshToken() {
+    if (!this.tokenData.needsRefresh()) {
+      return;
+    }
+    this.tokenData = synchronizedRequestToken();
+  }
+
```
mkistler: I don't think you need this function at all -- we should call `requestToken` directly when we need to refresh the token.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -211,12 +236,26 @@ public void authenticate(Builder builder) {
   public String getToken() {
     String token;
 
-    if (this.tokenData == null || !this.tokenData.isTokenValid()) {
-      // request new token
-      this.tokenData = requestToken();
+    if (this.tokenData == null) {
+      this.tokenData = synchronizedRequestToken();
+    } else if (this.tokenData.needsRefresh()) {
```
padamstx: Not to pile on here too much, but the logic might work out better if this second check was to detect that the tokenData represents a totally expired access token, in which case you'd just want to call synchronizedRequestToken() just like in the first check (for a null tokenData).
Then the third check would be to check if we need to refresh (even though the current token is valid but within the refresh window).   Just my $.02 :)

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -73,6 +77,24 @@ public boolean isTokenValid() {
...
 
-    return currentTime < refreshTime;
+    if (currentTime > refreshTime) {
+      // Advance expiration time by one minute.
+      this.expiration += 60;
+
+      return true;
+    }
+
+    return false;
```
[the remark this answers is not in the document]
lpatino10: Good call, I think this was a holdover from when we first implemented it. Making that change now.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
lpatino10: All these changes make it way cleaner. Thank you!

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
lpatino10: I just pushed a commit that addresses Mike's latest comments and I think that should take care of your second point about the `getToken()` logic.

I also did tweak the `IamToken` class, but I didn't break it out like we do with CP4D, mostly because I was a bit worried about moving public methods on `IamToken` to get fields in case it was a breaking change. I added the calculation logic to the constructor and that's working the same.

src/main/java/com/ibm/cloud/sdk/core/security/Cp4dToken.java
```diff
@@ -19,21 +19,22 @@
...
+  private long expirationTimeInMillis;
+  private long refreshTimeInMillis;
 
   /**
    * This ctor is used to store a user-managed access token which will never expire.
    * @param accessToken the user-managed access token
    */
   public Cp4dToken(String accessToken) {
     this.accessToken = accessToken;
-    this.expirationTimeInMillis = -1;
```
mkistler: If you don't set `expirationTimeInMillis` here, I assume it will be zero.  So won't `isTokenValid()` always return `false` for this token?  Is that what we want?

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -51,28 +52,51 @@ public Long getExpiration() {
...
+   * @return true iff the current access token is invalid or past the expiration buffer
    */
   @Override
-  public boolean isTokenValid() {
-    if (getAccessToken() == null || getExpiresIn() == null || getExpiration() == null) {
-      return false;
+  public synchronized boolean needsRefresh() {
+    long currentTime = System.currentTimeMillis() / 1000;
+
+    if (refreshTime == null || currentTime > refreshTime) {
```
mkistler: I think we need to be clear about what `refreshTime == null` means.  Does it mean "this token needs to be refreshed immediately" or does it mean "never refresh this token"?  I would have thought it was the latter, but here it seems to be taken to mean the former.

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -51,28 +52,51 @@ public Long getExpiration() {
     return expiration;
   }
 
+  IamToken() {
```
mkistler: Is this a constructor?  That takes no arguments? This is really confusing!

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -212,11 +225,20 @@ public String getToken() {
...
-      // request new token
-      this.tokenData = requestToken();
+      synchronizedRequestToken();
+    } else if (this.tokenData.needsRefresh()) {
+
+      // Kick off background task to refresh token.
+      Thread updateTokenCall = new Thread(new Runnable() {
+        @Override
+        public void run() {
+          synchronizedRequestToken();
```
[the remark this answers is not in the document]
lpatino10: Good call... I think I was worried about multiple threads trying to call that but the refresh time will be updated in `needsRefresh()` (which is synchronized) so that actually shouldn't happen. Changing now.

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
lpatino10: Yeah, I was adding package-private constructor that's called when we deserialize an IAM response. It does look a bit weird though. I can just make it private.

src/main/java/com/ibm/cloud/sdk/core/security/Cp4dToken.java
lpatino10: Nope, definitely a mistake there. I'll go ahead and set it to -1 like the refresh time to specify that it never expires and update the `isTokenValid()` logic accordingly.

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
lpatino10: Looking at it again maybe it should mean the former.

Originally I was thinking that if the `refreshTime` wasn't set, we were parsing some sort of invalid IAM response that was missing fields. It does seem possible though that there just wouldn't be `expires_in` or `expiration` fields in the response to indicate that it doesn't expire at all.

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -51,28 +52,51 @@ public Long getExpiration() {
...
 
+  private IamToken() {
+    if (getExpiresIn() != null && getExpiration() != null) {
+      Double fractionOfTimeToLive = 0.8;
+      Long timeToLive = getExpiresIn();
+      Long expirationTime = getExpiration();
+      this.refreshTime = expirationTime - (long) (timeToLive * (1.0 - fractionOfTimeToLive));
+    }
+  }
+
```
mkistler: It turns out that this constructor is called from GSON as part of the processing the response from the IAM token server.  But the way GSON works is that it will construct an "empty object", and then fill in the fields.  So when this constructor runs, all the fields are `null`, so `this.refreshTime` is never set.

I think what you should delete this constructor and put the logic for calculating `refreshTime` into `needsRefresh` below.

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -51,28 +52,51 @@ public Long getExpiration() {
...
+   *
+   * @return true iff the current access token is invalid or past the expiration buffer
    */
   @Override
-  public boolean isTokenValid() {
-    if (getAccessToken() == null || getExpiresIn() == null || getExpiration() == null) {
-      return false;
+  public synchronized boolean needsRefresh() {
+    long currentTime = System.currentTimeMillis() / 1000;
+
```
mkistler: Add the logic to calculate `refreshTime` here if `this.refreshTime == null`.  If you always set `refreshTime` to a non-null value, it will only be executed once -- so still efficient.

lpatino10:
I just pushed some more changes to hopefully resolve the latest comments.

I added a `Clock` class to wrap the calls to `System.currentTimeMillis()` to make testing easier. With this, I wrote tests for both the IAM and CP4D authenticators to validate the refresh logic, so getting those working has me feeling much more confident.

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -52,27 +54,46 @@ public Long getExpiration() {
...
+   * token expiring before the request could be made, so this method will return true
+   * if the current time is within that buffer.
    *
    * The buffer will be a fraction of the total TTL. Using 80%.
    *
-   * @return true iff the current access token is valid and not expired
+   * This method also updates the expiration time if it determines the token needs refreshed to prevent other
+   * threads from making multiple refresh calls.
+   *
+   * @return true iff the current access token is invalid or past the expiration buffer
```
mkistler: I think you can use the exact language from `Cp4dToken` to describe the `@return`:
>    * @return true if token is invalid or past the refresh time, false otherwise

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -52,27 +54,46 @@ public Long getExpiration() {
...
-    return currentTime < refreshTime;
+  /**
+   * Check if the currently stored access token is valid. This is different from the needsRefresh method in that it
+   * uses the actual TTL to calculate the expiration, rather than just a fraction.
+   *
+   * @return true iff is the current access token is not expired
+   */
+  @Override
+  public boolean isTokenValid() {
+    return Clock.getCurrentTimeInSeconds() < this.expiration;
```
mkistler: It's a little strange that we use `this.expiration` here, rather than `getExpiration()`, as we do in `needsRefresh`, and that we don't check for null here where we did in `needsRefresh`.  So could this property be null or not?  We seem to be schizophrenic on this.

For consistency, I think this line should be:
```
return this.expiration == null || Clock.getCurrentTimeInSeconds() < this.expiration;
```

So a null expiration time means "never expires".  I actually don't think this will happen, but again, for consistency, I think we need to handle it here.

PR #77, opened 2020-04-30
fix: trailing slash in request url & remove whitespace when parsing external config properties

jorge-ibm:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1738, https://github.ibm.com/arf/planning-sdk-squad/issues/1742

Changes/fixes:
- Empty path segments `""` are no longer added to the request url to prevent the java request url builder library from adding a `/` in place of the empty path segment
- When parsing a credentials file (vcap or environment file), the credential's key/value pairs are trimmed of whitespace

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
@@ -286,8 +286,8 @@ public final class CredentialUtils {
       Map<String, String> props = new HashMap<>();
       serviceName = serviceName.toUpperCase().replaceAll("-", "_") + "_";
       for (Map.Entry<String, String> entry : env.entrySet()) {
-        String key = entry.getKey();
-        String value = entry.getValue();
+        String key = entry.getKey().trim();
+        String value = entry.getValue().trim();
 
         if (key.startsWith(serviceName)) {
           String credentialName = key.substring(serviceName.length());
@@ -340,6 +340,7 @@ public final class CredentialUtils {
    */
   private static void addToMap(Map<String, String> map, String key, String value) {
     if (StringUtils.isNotEmpty(value)) {
+      value = value.trim();
       map.put(key, value);
     }
   }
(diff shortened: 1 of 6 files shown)


src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
```diff
@@ -340,6 +340,7 @@ protected static VcapService getVcapServiceEntry(String serviceName) {
    */
   private static void addToMap(Map<String, String> map, String key, String value) {
     if (StringUtils.isNotEmpty(value)) {
+      value = value.trim();
```
jorge-ibm, line 343: @padamstx - I went ahead and added this when parsing vcap creds as well.

src/test/java/com/ibm/cloud/sdk/core/test/service/RequestBuilderTest.java
```diff
@@ -286,6 +286,49 @@ public void testConstructHttpUrlGood() {
...
+    assertEquals("https://myserver.com/testservice/api", url.toString());
+  }
+
+  @Test
+  public void testConstructHttpUrlEmptyPathWParams() {
+    String[] pathSegments = { "" };
+    String[] pathParameters = { "param1", "param2" };
+    HttpUrl url = RequestBuilder.constructHttpUrl("https://myserver.com/testservice/api", pathSegments, pathParameters);
+    assertNotNull(url);
+    assertEquals("https://myserver.com/testservice/api/param1", url.toString());
```
padamstx, line 311: Could you check to see the result of this test BEFORE your fix was applied?
I'm sort of wondering whether we'd end up with a url of `https://myserver.com/testservice/api/param1/param2`.
I can imagine a path like "/{param1}/{param2}" which would result in "" within the pathSegments list, but {"param1", "param2" in the pathParameters list (I think).   And in that case I think we'd expect the url to be `https://myserver.com/testservice/api/param1/param2`, wouldn't we?
I realize this might be a corner case, but I can't help myself from always bringing those up :)

jorge-ibm: I just tested this without my changes and the result is still the same ``https://myserver.com/testservice/api/param1``. The `/` most likely got added by the empty path segment, then the library saw that param1 didn't need a a `/` in front of it

PR #79, opened 2020-05-08
feat: support service configuration via system properties

jorge-ibm:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1762

Changes:
- Support java system properties as an alternative to environment variables for config properties
- Add unit tests

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
@@ -23,6 +23,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -57,6 +58,7 @@ public final class CredentialUtils {
    * 1) Credential file
    * 2) Environment variables
    * 3) VCAP_SERVICES
+   * 4) System properties
    * The properties are returned in a Map.
    *
    * @param serviceName the name of the service. When searching for service-related properties in the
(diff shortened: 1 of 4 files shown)


src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
```diff
@@ -214,6 +221,10 @@ protected static VcapService getVcapServiceEntry(String serviceName) {
       files.add(new File(userSpecifiedPath));
     }
 
+    if (StringUtils.isNotEmpty(userSpecifiedSystemProp)) {
```
jorge-ibm, line 224: I couldn't move this to the end because unit tests would fail due to them finding other credential files in the java-core root directory

padamstx: yeah, I think this is fine.  If the user sets IBM_CREDENTIALS_FILE as an env var OR a system property, i think the intent is clear that they want to specify the custom filename as an override to any of the standard names we may look for.

src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java
```diff
@@ -302,6 +313,41 @@ protected static VcapService getVcapServiceEntry(String serviceName) {
...
+    Properties systemProperties = System.getProperties();
+
+    // Extract the properties related to the specified service and populate the
+    // result Map.
+    if (systemProperties != null) {
+      Map<String, String> props = new HashMap<>();
+      serviceName = serviceName.toUpperCase().replaceAll("-", "_") + "_";
+
+      for (Map.Entry<Object, Object> entry : systemProperties.entrySet()) {
+        String key = entry.getKey() == null ? "" : String.valueOf(entry.getKey()).trim();
```
padamstx, line 335: I think it would be ok to just check if the entry's key or value is null, and if so, just "continue" (i.e. skip that entry altogether).

PR #83, opened 2020-07-23
fix: surface ServiceResponseError outside of RuntimeError

christiancompton:
This PR will allow ServiceResponseErrors to be returned directly, as opposed to in the `causedBy:` of a RuntimeException.

### DIFF
--- a/src/test/java/com/ibm/cloud/sdk/core/test/security/Cp4dAuthenticatorTest.java
+++ b/src/test/java/com/ibm/cloud/sdk/core/test/security/Cp4dAuthenticatorTest.java
@@ -285,10 +285,33 @@ public class Cp4dAuthenticatorTest extends BaseServiceUnitTest {
     try {
       authenticator.authenticate(requestBuilder);
       fail("Expected authenticate() to result in exception!");
+    } catch (ServiceResponseException excp) {
+      assertTrue(excp instanceof ServiceResponseException);
+    } catch (Throwable t) {
+      fail("Expected ServiceResponseException, not " + t.getClass().getSimpleName());
+    }
+  }
+
+  @Test
+  public void testApiErrorResponse() throws Throwable {
+    server.enqueue(jsonResponse("{'}"));
+
+    // Mock current time to ensure the token is valid.
+    PowerMockito.mockStatic(Clock.class);
+    PowerMockito.when(Clock.getCurrentTimeInSeconds()).thenReturn((long) 100);
+
+    CloudPakForDataAuthenticator authenticator = new CloudPakForDataAuthenticator(url, testUsername, testPassword);
+
+    Request.Builder requestBuilder = new Request.Builder().url("https://test.com");
+
+    // Calling authenticate should result in an exception.
+    try {
+      authenticator.authenticate(requestBuilder);
+      fail("Expected authenticate() to result in exception!");
     } catch (RuntimeException excp) {
       Throwable causedBy = excp.getCause();
       assertNotNull(causedBy);
-      assertTrue(causedBy instanceof ServiceResponseException);
+      assertTrue(causedBy instanceof IllegalStateException);
     } catch (Throwable t) {
       fail("Expected RuntimeException, not " + t.getClass().getSimpleName());
     }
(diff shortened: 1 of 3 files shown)


src/test/java/com/ibm/cloud/sdk/core/test/security/Cp4dAuthenticatorTest.java
```diff
@@ -281,14 +281,37 @@ public void testApiErrorBadRequest() throws Throwable {
...
+
     // Calling authenticate should result in an exception.
     try {
       authenticator.authenticate(requestBuilder);
       fail("Expected authenticate() to result in exception!");
     } catch (RuntimeException excp) {
       Throwable causedBy = excp.getCause();
       assertNotNull(causedBy);
-      assertTrue(causedBy instanceof ServiceResponseException);
+      assertFalse(causedBy instanceof ServiceResponseException);
```
padamstx, line 314: I think it would be better to check for the specific exception type that we'd expect in this "bad json response" scenario (i.e. a JsonParseException), rather than just make sure it's not a ServiceResponseException.

src/test/java/com/ibm/cloud/sdk/core/test/security/Cp4dAuthenticatorTest.java
```diff
@@ -281,14 +281,37 @@ public void testApiErrorBadRequest() throws Throwable {
 
     Request.Builder requestBuilder = new Request.Builder().url("https://test.com");
 
+    // Calling authenticate should result in an exception.
+    try {
+      authenticator.authenticate(requestBuilder);
+      fail("Expected authenticate() to result in exception!");
+    } catch (RuntimeException excp) {
```
padamstx, line 288:
```suggestion
    } catch (ServiceResponseException excp) {
```
You could just catch the ServiceResponseException directly and call that a success, while anything else (the Throwable catch clause) would be an error.

src/test/java/com/ibm/cloud/sdk/core/test/security/IamAuthenticatorTest.java
```diff
@@ -338,6 +338,31 @@ public void testApiErrorBadRequest() throws Throwable {
...
     IamAuthenticator authenticator = new IamAuthenticator(API_KEY);
+    authenticator.setURL(url);
+
+    Request.Builder requestBuilder = new Request.Builder().url("https://test.com");
+
+    // Calling authenticate should result in an exception.
+    try {
+      authenticator.authenticate(requestBuilder);
+      fail("Expected authenticate() to result in exception!");
+    } catch (RuntimeException excp) {
```
padamstx, line 349:
```suggestion
    } catch (ServiceResponseException excp) {
```
Same comment as above.

src/test/java/com/ibm/cloud/sdk/core/test/security/IamAuthenticatorTest.java
```diff
@@ -348,7 +373,8 @@ public void testApiErrorBadRequest() throws Throwable {
     } catch (RuntimeException excp) {
       Throwable causedBy = excp.getCause();
       assertNotNull(causedBy);
-      assertTrue(causedBy instanceof ServiceResponseException);
+      assertFalse(causedBy instanceof ServiceResponseException);
```
padamstx, line 376:
```suggestion
      assertTrue(causedBy instanceof JsonParseException);
```

src/test/java/com/ibm/cloud/sdk/core/test/security/IamAuthenticatorTest.java
```diff
@@ -348,7 +373,8 @@ public void testApiErrorBadRequest() throws Throwable {
...
+      assertFalse(excp instanceof ServiceResponseException);
```
padamstx, line 377:
```suggestion
```
I don't think this is needed if we correctly test the happy path.

src/test/java/com/ibm/cloud/sdk/core/test/security/Cp4dAuthenticatorTest.java
christiancompton: This should remain `RuntimeExcpetion` here, since the underlying error is IllegalStateException from the bad JSON. The ServiceResponse Exception changes have been made in both the `testApiErrorBadRequest()` locations where it is now surfaced.

padamstx: I was just testing you :)  HAHAHA

PR #85, opened 2020-07-28
BREAKING CHANGE: upgrade okhttp3 dependency to 3.14.9

jorge-ibm:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1957

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: java
 dist: trusty
 jdk:
-- openjdk7
-- oraclejdk8
+- openjdk8
 
 notifications:
   email: true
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: java
 dist: trusty
 jdk:
-- openjdk7
+- openjdk8
 - oraclejdk8
```
padamstx, line 5:
```suggestion
```
I don't think we need both flavors of java 8.  Let's just pick "openjdk8" since that is consistent with the java template repo and other SDK projects, etc.

PR #86, opened 2020-07-31
Re-apply previous commit with new semrel-friendly commit message

padamstx:
The net effect of this PR is to simply modify the commit message associated with
@jorge-ibm 's original commit that upgrades our dependency on okhttp3.


padamstx:
The fact that there are no net file changes to display for this PR tells me that I was able to pull it off :)

PR #88, opened 2020-08-06
fix: avoid NPE when constructing dynamic model

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/2060

This commit fixes the scenario where an NPE could occur when
constructing an instance of a dynamic model that contains no
dynamic properties.  We were simply missing a null check
prior to using the generated builder class' dynamicProperties field.


src/test/java/com/ibm/cloud/sdk/core/test/model/DiscriminatorSerializationTest.java
```diff
@@ -39,9 +39,9 @@
 import com.ibm.cloud.sdk.core.util.GsonSingleton;
```
padamstx, line 39: I changed this test class to use Builder classes when constructing the Animal-related model instances.
When Builder classes were introduced for dynamic models, it was done such that we are still generating the 
setter methods that we were previously generating for dynamic models as a way to maintain compatibility.
And that worked as we didn't "need" to change our dynamic model-related tests here in the core.
But that allowed this NPE bug to slip by :(

src/test/java/com/ibm/cloud/sdk/core/test/model/generated/Animal.java
```diff
@@ -1,5 +1,5 @@
 /*
- * (C) Copyright IBM Corp. 2019.
```
padamstx, line 2: I copied in new versions of the Animal-related classes from the generator test snapshots in order to get the Builder classes that were previously introduced for dynamic models.

PR #89, opened 2020-08-07
feat: add ratelimiter to allow transparent retry of 429

dpittner:
What: adds a basic rate-limiter implementation 

 - had to make authenticator protected, as retrying requests may require refreshing auth. 
 - Will cause the calling thread to sleep, question is if long term sdk-core should use dispatcher and bulk-head pattern.

- Usage example included in the test: https://github.com/IBM/java-sdk-core/compare/master...dpittner:rateLimiter?expand=1#diff-7fdf4f8677750de0b55ff0c839b7a32aR51


dpittner:
@padamstx sure let me know, would be great if we could discuss anything remaining in todays call, such that we can get this in timely.

PR #92, opened 2020-09-11
feat: add method RequestBuilder.resolveRequestUrl

padamstx:
This commit adds method resolveRequestURL to the RequestBuilder class
as a replacement for the constructHttpUrl() method.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,7 +42,7 @@
         <compiler-plugin-version>3.8.0</compiler-plugin-version>
         <okhttp3-version>3.14.9</okhttp3-version>
         <logback-version>1.2.3</logback-version>
-        <guava-version>27.1-android</guava-version>
+        <guava-version>29.0-jre</guava-version>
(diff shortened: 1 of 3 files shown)


padamstx:
Removed reviewers because I noticed a couple things that need to be fixed.

pom.xml
```diff
@@ -42,7 +42,7 @@
         <compiler-plugin-version>3.8.0</compiler-plugin-version>
         <okhttp3-version>3.14.9</okhttp3-version>
         <logback-version>1.2.3</logback-version>
-        <guava-version>27.1-android</guava-version>
```
jorge-ibm, line 45: 🤔

jorge-ibm (approved):
LGTM

pom.xml
padamstx: while the old version # refers to android, that version supported java 7+, but we no longer need java 7 support so I just moved us up to the latest.

PR #93, opened 2020-09-16
feat(IAM Authenticator): add support for optional 'scope' property

jorge-ibm:
The IAM /identity/token (get token) operation supports an optional "scope" form parameter which can be used to obtain an IAM access token having a limited set of scopes in this it can be used.

This commit adds support for the "Scope" field within the IamAuthenticator class. If set, the value of this field is sent as the "scope" form param in the "get token" request body when obtaining an access token.

ref: https://github.ibm.com/arf/planning-sdk-squad/issues/2167

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/security/IamAuthenticator.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/security/IamAuthenticator.java
@@ -79,6 +79,30 @@ public class IamAuthenticator extends TokenRequestBasedAuthenticator<IamToken, I
   }
 
   /**
+   * Constructs an IamAuthenticator with all properties.
+   *
+   * @param apikey
+   *          the apikey to be used when retrieving the access token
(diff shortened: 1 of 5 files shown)


src/main/java/com/ibm/cloud/sdk/core/security/IamAuthenticator.java
```diff
@@ -73,7 +75,7 @@ public IamAuthenticator(String apikey) {
    */
   public IamAuthenticator(String apikey, String url, String clientId, String clientSecret,
```
padamstx, line 76: I think for Java, we could add a new public ctor that included the new scope parameter since Java allows overloaded functions.

padamstx (changes requested):
Looks good, I just had one suggestion to add a new public ctor that supports the scope field.

padamstx (approved):
LGTM

PR #94, opened 2020-09-25
feat: add support for gzip request body compression and upgrade okhttp version

jorge-ibm:
This PR adds support for compressing request bodies using gzip. A new ``enableGzipCompression`` method was added as part of the BaseService class, and is used to add the ``GzipRequestInterceptor`` to the instance of the Base Service's http client. This new interceptor sets the ``Content-Encoding`` of the request body to ``gzip``, if no ``Content-Encoding`` header exists and then compresses the request body.

**The version of okhttp was also upgraded to v4.9.0**

ref: https://github.ibm.com/arf/planning-sdk-squad/issues/2181

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
@@ -265,6 +265,30 @@ public class HttpClientSingleton {
   }
 
   /**
+   * Sets a new list of interceptors for the specified {@link OkHttpClient} instance by removing the specified
+   * interceptor and returns a new instance with the interceptors configured as requested.
+   *
+   * @param client the {@link OkHttpClient} instance to set the proxy authenticator on
+   * @param interceptorToRemove the class name of the interceptor to remove
+   * @return the new {@link OkHttpClient} instance with the new list of interceptors
+   */
+  private OkHttpClient reconfigureClientInterceptors(OkHttpClient client, String interceptorToRemove) {
+    OkHttpClient.Builder builder = client.newBuilder();
+
+    if (!builder.interceptors().isEmpty()) {
+      for (int i = 0; i < builder.interceptors().size(); i++) {
+        String currentInterceptor = builder.interceptors().get(i).getClass().getSimpleName();
+        if (currentInterceptor.equals(interceptorToRemove)) {
+          LOG.log(Level.INFO, "Removing interceptor" + currentInterceptor + " from http client instance.");
+          builder.interceptors().remove(i);
+        }
+      }
+    }
+
+    return builder.build();
+  }
+
+  /**
    * Creates a new {@link OkHttpClient} instance with a new {@link ServiceCookieJar}
    * and a default configuration.
    *
@@ -318,6 +342,7 @@ public class HttpClientSingleton {
                         , options.getMaxRetries()))
                 .build();
       }
+      client = reconfigureClientInterceptors(client, "GzipRequestInterceptor");
       if (options.shouldEnableGzipCompression()) {
         client = client.newBuilder()
                 .addInterceptor(new GzipRequestInterceptor())
--- a/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
@@ -15,6 +15,7 @@ package com.ibm.cloud.sdk.core.http;
 
 import com.ibm.cloud.sdk.core.http.HttpConfigOptions.LoggingLevel;
 import com.ibm.cloud.sdk.core.http.ratelimit.RateLimitInterceptor;
+import com.ibm.cloud.sdk.core.http.gzip.GzipRequestInterceptor;
 import com.ibm.cloud.sdk.core.service.BaseService;
 import com.ibm.cloud.sdk.core.service.security.DelegatingSSLSocketFactory;
 import okhttp3.Authenticator;
@@ -315,6 +316,11 @@ public class HttpClientSingleton {
                         options.getAuthenticator()
                         , options.getDefaultRetryInterval()
                         , options.getMaxRetries()))
+                .build();
+      }
+      if (options.shouldEnableGzipCompression()) {
+        client = client.newBuilder()
+                .addInterceptor(new GzipRequestInterceptor())
                 .build();
       }
     }
--- a/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
@@ -20,6 +20,7 @@ import com.ibm.cloud.sdk.core.service.BaseService;
 import com.ibm.cloud.sdk.core.service.security.DelegatingSSLSocketFactory;
 import okhttp3.Authenticator;
 import okhttp3.ConnectionSpec;
+import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 import okhttp3.OkHttpClient.Builder;
 import okhttp3.TlsVersion;
@@ -43,6 +44,7 @@ import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
 import java.util.Arrays;
+import java.util.Iterator;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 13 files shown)


jorge-ibm:
@ricellis - can you please try out the new compression feature using this branch, building the core locally, & let me know if you run into any issues? It can be enabled by calling ``service.enableGzipCompression(true)`` from the service instance or by setting an environment property ``ENABLE_GZIP=true``. Thanks

jorge-ibm:
I think this PR is ready to go, but I'm trying to do a test against a live service, so far 2/2 services have not accepted a gzip compressed request body (vulnerability advisor + config governance). I'm hoping to find a service that'll accept it to verify..

padamstx:
> I think this PR is ready to go, but I'm trying to do a test against a live service, so far 2/2 services have not accepted a gzip compressed request body (vulnerability advisor + config governance). I'm hoping to find a service that'll accept it to verify..

I would *hope* that the IAM Identity service allows it.  If you want to try that out, you could pull the "add-iam-identity" feature branch in the platform-services Java SDK.   That service has not officially been merged into the project, but that feature branch should work just fine from an integration test standpoint.

padamstx:
> @ricellis - can you please try out the new compression feature using this branch, building the core locally, & let me know if you run into any issues? It can be enabled by calling `service.enableGzipCompression(true)` from the service instance or by setting an environment property `ENABLE_GZIP=true`. Thanks

Actually, I think one would have to use a property name of the form `<service_name>_ENABLE_GZIP`, similar to other config properties.

src/main/java/com/ibm/cloud/sdk/core/http/gzip/GzipRequestInterceptor.java
```diff
@@ -0,0 +1,45 @@
+package com.ibm.cloud.sdk.core.http.gzip;
```
padamstx, line 1: Should probably add a copyright header.

src/main/java/com/ibm/cloud/sdk/core/http/gzip/GzipRequestInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+import okhttp3.Interceptor;
+import okhttp3.MediaType;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okio.BufferedSink;
+import okio.GzipSink;
+import okio.Okio;
+
+/** This interceptor compresses the HTTP request body. Many webservers can't handle this! */
```
padamstx, line 13: Should probably add a reference to where this class was copied from:
https://square.github.io/okhttp/interceptors/#rewriting-requests

src/test/java/com/ibm/cloud/sdk/core/test/http/gzip/GzipTest.java
```diff
@@ -0,0 +1,824 @@
...
+
+        // validate response
+        Response<TestModel> response = service.testMethod(builder).execute();
+        assertNotNull(response);
+        TestModel responseObj = response.getResult();
+        assertNotNull(responseObj);
+        assertEquals(201, response.getStatusCode());
+        assertEquals("awesome", responseObj.getSuccess());
+    
+        // Verify the request was indeed compressed, content encoding, & content length
```
padamstx, line 155:
```suggestion
        // Verify the request was indeed uncompressed, content encoding, & content length
```
???

padamstx (changes requested):
This PR looks good.  I had a couple minor comments re: a copyright prolog, etc.   
You seemed to cover all the various request use cases.  
I realize that the focus of this PR is on the request path, and the handling of gzip stuff on the response path is automatically handled by okhttp.  However, I think it would be good to slightly modify a couple of the new tests to also make sure that the request has the `Accept-Encoding: gzip` header set on it.   This is essentially where the client is telling the server that it can handle a gzipped response, and if the server decides to compress the response it would include the `Content-Encoding: gzip` header in the response.  Perhaps while doing this, you could also simulate a gzipped response or two?

ricellis:
>can you please try out the new compression feature using this branch, building the core locally, & let me know if you run into any issues?

I built this locally and was able to successfully enable the request body compression via:
`service.enableGzipCompression(true)` and `<service>_ENABLE_GZIP=true`

AFAICT it seems to be working as expected in that the `Content-Encoding: gzip` header is present and the body of the request is gzip compressed.

A couple of things I noticed:
1. If `Level.BODY` logging is enabled then the compressed bodies are omitted e.g.
    ```
    Sep 28, 2020 10:31:53 AM okhttp3.internal.platform.Platform log
    INFO: --> END PUT (binary -1-byte body omitted)
    ```
1. I wasn't able to disable compression once it had been enabled. For example:
    * if it had been enabled via the env var, then I wasn't able to override that by calling `client.enableGzipCompression(false)`
    * if I added `client.enableGzipCompression(true)` in our generated `Service.java` then calling `client.enableGzipCompression(false)` on the user-created instance didn't disable it.

I think we _might_ need either one or the other to work to facilitate client-side debugging.
If we can't log the body without it being decompressed then we might need the toggle to work so that we can ask a customer to turn off compression to collect a body log for debugging.
Alternatively if we can log the body some other way then it might be OK that it's a 1-way enablement, but in that case I do think it is a little strange to expose the method at all (i.e. there is no point letting a user call `enableGzipCompression` if it has no effect and the call in the generated ctor can't be reversed).

src/test/java/com/ibm/cloud/sdk/core/test/http/gzip/GzipTest.java
jorge-ibm: copy and paste mistake 🙂 . Good catch

jorge-ibm:
thanks @ricellis - I'll work on addling the ability to disable gzip once enabled.

src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
```diff
@@ -263,6 +264,31 @@ protected SSLSocket configureSocket(SSLSocket socket) throws IOException {
...
+   */
+  private OkHttpClient reconfigureClientInterceptors(OkHttpClient client, String interceptorToRemove) {
+    OkHttpClient.Builder builder = client.newBuilder();
+
+    if (!builder.interceptors().isEmpty()) {
+      for (int i = 0; i < builder.interceptors().size(); i++) {
+        String currentInterceptor = builder.interceptors().get(i).getClass().getSimpleName();
+        if (currentInterceptor.equals(interceptorToRemove)) {
+          LOG.log(Level.INFO, "Removing interceptor" + currentInterceptor + " from http client instance.");
+          System.out.println("Removing interceptor" + currentInterceptor + " from http client instance.");
```
padamstx, line 283: I don't think we need to log to the logger AND stdout, do we?

src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
```diff
@@ -317,6 +343,15 @@ public OkHttpClient configureClient(OkHttpClient client, HttpConfigOptions optio
                         , options.getMaxRetries()))
                 .build();
       }
+      if (options.shouldEnableGzipCompression()) {
```
padamstx, line 346: I know this is sort of pathological, but what would happen if the user called "enableGzipCompression(true)" twice?  :)

One way to harden this piece of code that either adds or removes the interceptor would be to unconditionally remove the Gzip interceptor, then if "shouldEnableGzipCompression()" is true, then add it.   That way, we're guaranteed to have the Gzip interceptor registered IFF shouldEnableGzipCompression() is true and it will be present in the list at most once.

padamstx (approved):
Looks good, save for a couple comments I added.   Approving now to avoid an extra review cycle.

src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
jorge-ibm: Nope, that was a remnant of my debugging. I'll remove it.

src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
```diff
@@ -263,6 +264,30 @@ protected SSLSocket configureSocket(SSLSocket socket) throws IOException {
...
+   *
+   * @param client the {@link OkHttpClient} instance to set the proxy authenticator on
+   * @param interceptorToRemove the class name of the interceptor to remove
+   * @return the new {@link OkHttpClient} instance with the new list of interceptors
+   */
+  private OkHttpClient reconfigureClientInterceptors(OkHttpClient client, String interceptorToRemove) {
+    OkHttpClient.Builder builder = client.newBuilder();
+
+    if (!builder.interceptors().isEmpty()) {
+      for (int i = 0; i < builder.interceptors().size(); i++) {
```
padamstx, line 279: I think you need to use an iterator to walk through the list rather than an integer index because you are potentially removing an item from the list as you are walking the list.   Iterator-based traversals allow you to remove an item safely.

padamstx (approved):
Looks good... just had the one comment about removing an interceptor from within an index-based loop.  I'll approve now to avoid another review.

padamstx (approved):
Looks good!

PR #100, opened 2020-11-05
feat: add new DateUtils utility class

padamstx:
Fixes arf/planning-sdk-squad#2289

This commit introduces a new class (DateUtils) which will
be used by the Java generator to properly format date and date-time
values when they are used as query parameters.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/DateDeserializer.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/DateDeserializer.java
@@ -19,8 +19,6 @@ import java.text.SimpleDateFormat;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -60,8 +58,6 @@ public class DateDeserializer implements JsonDeserializer<Date> {
 
   private final List<SimpleDateFormat> rfc822Formatters =
       Arrays.asList(rfc822DateFormatter, rfc822WithoutMsDateFormatter);
-
-  private static final Logger LOG = Logger.getLogger(DateDeserializer.class.getName());
 
   /*
    * (non-Javadoc)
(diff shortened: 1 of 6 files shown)


jorge-ibm (approved):
LGTM

src/main/java/com/ibm/cloud/sdk/core/util/DateDeserializer.java
```diff
@@ -113,8 +109,6 @@ public synchronized Date deserialize(JsonElement json, Type typeOfT, JsonDeseria
       return new Date(timeAsLong);
     }
 
-    LOG.log(Level.SEVERE, "Error parsing: " + dateAsString, e);
```
padamstx, line 116: While working on this issue, I discovered that we were just absorbing an exception caused by an incorrectly-formatted date/datetime value, so made this change to throw an exception in that case.

src/main/java/com/ibm/cloud/sdk/core/util/OpenAPIDateSerializer.java
```diff
@@ -0,0 +1,41 @@
+/**
```
padamstx, line 1: This class is not yet used by the generator, but will be used later when we address the formatting of model properties.

src/main/java/com/ibm/cloud/sdk/core/util/OpenAPIDateTimeSerializer.java
```diff
@@ -0,0 +1,41 @@
+/**
```
padamstx, line 1: This class is not yet used by the generator, but will be used later when we address the formatting of model properties.

src/test/java/com/ibm/cloud/sdk/core/util/DateDeserializerTest.java
```diff
@@ -17,14 +17,13 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 
+import java.text.SimpleDateFormat;
```
padamstx, line 20: Looks like an intellij user previously messed with this class...  @mkistler any idea who that might have been?  :)

PR #107, opened 2021-01-28
feat: wrap exceptions during client-side response processing

padamstx:
This commit introduces a new InvalidServiceResponseException class,
which will wrap any exception caught while processing the response
received from the server for an otherwise successful operation
invocation.  For example, an exception thrown by GSON due to a
JSON unmarshalling error will now be wrapped in an
InvalidServiceResponseException (a subclass of ServiceResponseException).

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2015, 2019.
+ * (C) Copyright IBM Corp. 2015, 2021.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
@@ -33,6 +33,7 @@ import com.ibm.cloud.sdk.core.service.exception.BadRequestException;
 import com.ibm.cloud.sdk.core.service.exception.ConflictException;
 import com.ibm.cloud.sdk.core.service.exception.ForbiddenException;
 import com.ibm.cloud.sdk.core.service.exception.InternalServerErrorException;
+import com.ibm.cloud.sdk.core.service.exception.InvalidServiceResponseException;
 import com.ibm.cloud.sdk.core.service.exception.NotFoundException;
 import com.ibm.cloud.sdk.core.service.exception.RequestTooLargeException;
 import com.ibm.cloud.sdk.core.service.exception.ServiceResponseException;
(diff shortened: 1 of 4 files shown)


src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
```diff
@@ -254,7 +255,7 @@ protected void setDefaultHeaders(final Request.Builder builder) {
    */
   protected final <T> ServiceCall<T> createServiceCall(final Request request, final ResponseConverter<T> converter) {
     final Call call = createCall(request);
-    return new WatsonServiceCall<>(call, converter);
+    return new IBMCloudSDKServiceCall<>(call, converter);
```
padamstx, line 258: One last (hopefully) reference to Watson in the java core :)

src/main/java/com/ibm/cloud/sdk/core/service/exception/ServiceResponseException.java
```diff
@@ -40,40 +40,62 @@
   private static final Type debuggingInfoType = new TypeToken<Map<String, Object>>() { }.getType();
 
   /** The status code. */
-  private final int statusCode;
+  private int statusCode;
```
padamstx, line 43: Had to remove final here because of the new way of initializing "this".

src/test/java/com/ibm/cloud/sdk/core/test/service/ResponseTest.java
```diff
@@ -144,11 +147,9 @@ public void setUp() throws Exception {
 
   /**
    * Test that all fields are populated when calling execute().
-   *
-   * @throws InterruptedException the interrupted exception
    */
   @Test
-  public void testExecuteTestModel() throws InterruptedException {
+  public void testExecuteTestModel() {
```
padamstx, line 152: For some reason, these extraneous throws clauses were bugging me :)

src/test/java/com/ibm/cloud/sdk/core/test/service/ResponseTest.java
```diff
@@ -166,74 +176,124 @@ public void testExecuteTestModel() throws InterruptedException {
   public void testEnqueue() throws InterruptedException {
     server.enqueue(new MockResponse().setBody(testResponseBody1));
 
+    final Map<String, Object> results = new HashMap<>();
```
padamstx, line 179: The only change to this test method is how the results are now saved in the map.
I did this so that the callback can record the results of the test, but still allow us to validate those results in the main thread, since that's where we'll want to fail an assertXXX() so that JUnit will properly fail the test.

mkistler (approved):
Looks good! 👍

PR #108, opened 2021-01-31
feat(authentication): support cp4d /v1/authorize operation

padamstx:
The main focus of this change is to enhance the CloudPakForDataAuthentictor
implementation to support the official "POST /v1/authorize" operation for
obtaining an access token.   This includes support for using either a password
or an apikey when obtaining the access token from the token service.

In addition, a new Builder class was introduced which provides a more user-friendly
way to construct instances of CloudPakForDataAuthenticator.  The "fromConfiguration"
method was also added to replace the (now deprecated) map-based ctor.

In order to keep the IamAuthenticator class in sync with CloudPakForDataAuthenticator,
the Builder class and fromConfiguration() method were also added to IamAuthenticator.
In the case of both authenticators, the existing set of constructors are still supported
but are now deprecated and will be removed in the next major release of the Java core.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ config
 .secrets.baseline
 .vscode/
 package-lock.json
+*.env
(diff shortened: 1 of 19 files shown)


mkistler (approved):
Looks good! 👍

rmkeezer (approved):
This looks good! If the Java generator doesn't need any related changes to this now using the `Builder` class, then this should be good to go. And I assume the previous non-builder method still works, it's just deprecated now correct?

padamstx:
> This looks good! If the Java generator doesn't need any related changes to this now using the `Builder` class, then this should be good to go. And I assume the previous non-builder method still works, it's just deprecated now correct?

correct, we still support all the ctors we supported before

PR #110, opened 2021-02-07
fix(build): publish artifacts directly to maven central repository

padamstx:
This commit modifies the build configuration so that build artifacts
are published to the Maven Central repository (via the ossrh
staging repo).  Previously, build artifacts were published to
the bintray/jcenter repository, then mirrored to Maven Central.
But, with the imminent demise of bintray/jcenter, we need to
publish directly to Maven Central.  Going forward, published jars
will be available only on Maven Central.


.travis.yml
```diff
@@ -1,17 +1,13 @@
+---
```
padamstx, line 1: The .travis.yml file is now virtually the same as the one from the platform-services SDK.

mkistler (approved):
Looks good! 👍

dpopp07 (approved):
Looks good 👍

PR #111, opened 2021-02-10
fix(build): main migration

christiancompton:
This PR upgrades automation and documentation to use the new `main` branch.


padamstx (approved):
Change look good, but please change the commit message to use `fix(build)` instead of `chore` to force a new patch level to be built, as a way to exercise all the build tooling with the new branch name, etc.

PR #114, opened 2021-03-01
test: fix iam live test

padamstx:
This PR updates the IAM live server test to fix the comments that described the required config properties, plus it adds a test that verifies that a refresh token can be obtained using the authenticator.


padamstx:
@dpopp07 @rmkeezer Dustin, Matthew... I added you both as reviewers so you're aware of the IAM live server test that's part of the Java core.   Perhaps we could add something similar for Node and Python.

dpopp07 (approved):
Looks good! 👍

rmkeezer (approved):
LGTM!

PR #116, opened 2021-03-07
Add getQueryParam utility method to support pagination

mkistler:
This PR adds a small utility method to extract the value of a query parameter from a URL string. This will be used in the pagination support for Java.


padamstx (approved):
LGTM

PR #117, opened 2021-04-01
fix(deps): update dependencies to avoid vulnerabilities

padamstx:
Updates the guava dependency to avoid vulnerabilities, plus a few other deps.
Oh, and I also fixed a few javadoc warnings.

### DIFF
--- a/build/checkstyle.xml
+++ b/build/checkstyle.xml
@@ -4,36 +4,35 @@
 
 <module name="Checker">
     <module name="SuppressionFilter">
-        <property name="file"
-            value="build/checkstyle-suppressions.xml" />
+        <property name="file" value="build/checkstyle-suppressions.xml"/>
     </module>
 
-    <module name="NewlineAtEndOfFile" />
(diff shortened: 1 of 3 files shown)


build/checkstyle.xml
```diff
@@ -4,36 +4,35 @@
```
padamstx, line 4: Did a reformat on this file... there is really only 1 small change which was needed when upgrading to the latest checkstyle.

src/main/java/com/ibm/cloud/sdk/core/util/UrlHelper.java
```diff
@@ -22,6 +22,9 @@ private UrlHelper() {
 
     /**
      * Returns the value of query parameter `param` from urlStr, or null when not found.
+     * @param urlStr the URL string containing the query param to retrieve
```
padamstx, line 25: Somehow, these javadoc warnings previously evaded our intricate testing framework LOL

mkistler (approved):
Looks good! 👍

PR #118, opened 2021-04-20
test: adds tests re: missing/empty empty operation responses

padamstx:
This PR contains a few new test cases involving empty or missing operation responses.  This was mainly done to confirm the behavior of the Java core (and by extension Gson) in these situations.


mkistler (approved):
Looks good! 👍

PR #122, opened 2021-05-17
feat: add `BaseService.constructServiceURL` method

ajhynes7:
This static method can be used by generated code.


padamstx (approved):
Changes look good.   However, let's hold off on merging this until we finish the end-to-end scenario for Java (i.e. generator changes tested via systest framework... we can discuss the details here).

PR #125, opened 2021-06-02
test: add tests involving Host header

padamstx:
This commit adds tests to verify that the "Host" request
header can be explicitly set in the following ways:
- As part of a service request (request builder)
- As a custom header while constructing an IAM Authenticator instance
- As a custom header while constructing a CP4D Authenticator instance


dpopp07 (approved):
Looks good to me!

PR #126, opened 2021-06-07
fix: change method name to `constructServiceUrl`

ajhynes7:
This is consistent with the naming convention of `setServiceUrl`.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
@@ -241,9 +241,7 @@ public abstract class BaseService {
   }
 
   /**
-   * @deprecated
-   * Deprecated in favor of `constructServiceUrl`.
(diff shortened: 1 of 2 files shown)


src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
```diff
@@ -240,6 +240,20 @@ public static String constructServiceURL(
     return formattedUrl;
   }
 
+  /**
+   * @deprecated
+   * Deprecated in favor of `constructServiceUrl`.
+   * This naming convention is consistent with `setServiceUrl`.
```
padamstx, line 246:
```suggestion
   * @deprecated use constructServiceUrl() instead.
```
Let's just keep the message simple like other deprecated tags.

padamstx (approved):
Looks good, but I had just one minor comment about the @deprecated javadoc tag :)
I'll approve now assuming that is addressed.

PR #128, opened 2021-07-04
chore: Improve unit test coverage

Andras-Csanyi:
This PR contains 100% unit test coverage against `BooleanToStringTypeAdapter`.


pyrooka (approved):
Looks good to me!

Andras-Csanyi:
Let's hold off on this PR for a while.

padamstx:
@Andras-Csanyi I don't want to burst your bubble, however... :)   I did a quick search and it looks like we do not actually use the BooleanToStringTypeAdapter.  Perhaps at one time (before I joined the team) we did in fact register that type adapter with Gson, but it looks like we currently do not.  With that in mind, I think the best way to handle this one is to simply delete the class, rather than add testcases for it.  

Perhaps we can meet up on webex and take a look at the jacoco report to see where we can achieve the "biggest bang for the buck" in terms of improving test coverage.  For example, the jacoco report doesn't include any test coverage for the CloudPakForDataServiceAuthenticator despite the fact that it appears to be run during the "mvn test" phase.   Not sure why that is not contributing any test coverage data.   Also, there are probably a few other classes that are causing the coverage numbers to be dragged down.

padamstx (approved):
LGTM

PR #130, opened 2021-07-06
chore: improve unit test coverage of discriminatorbasedtypeadapter

Andras-Csanyi:
This PR improves unit test coverage of `DiscriminatorBasedTypeAdapterFactory`.

Coverage is improved to `92%` method and `68%` line coverage from `30%` method and `16%` line coverage.

I created a new package for test classes in the test `util` package to have them separated. The reason is that `1.7` java version is in the `pom.xml` (see the ticket on the team board about it) which doesn't allow to use static fields in nested classes. Once the version is bumped to `1.8` or higher these classes can be reorganized.


src/test/java/com/ibm/cloud/sdk/core/util/discriminator/AcceptPayloadWithoutDiscriminatorMapping.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.ibm.cloud.sdk.core.util.discriminator;
+
+public class AcceptPayloadWithoutDiscriminatorMapping extends StatusPayloadWithoutDiscriminatorMapping {
+
+}
```
pyrooka, line 17: I am not quite sure about the best style here, but I think the newline shouldn't be here.
```suggestion
public class AcceptPayloadWithoutDiscriminatorMapping extends StatusPayloadWithoutDiscriminatorMapping {
}
```

src/test/java/com/ibm/cloud/sdk/core/util/discriminator/AcceptPayloadWithoutDiscriminatorPropertyName.java
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.ibm.cloud.sdk.core.util.discriminator;
+
+
+public class AcceptPayloadWithoutDiscriminatorPropertyName extends StatusPayloadWithoutDiscriminatorPropertyName {
+
+}
```
pyrooka, line 18: Same as above.
```suggestion
public class AcceptPayloadWithoutDiscriminatorPropertyName extends  StatusPayloadWithoutDiscriminatorPropertyName {
}
```

src/test/java/com/ibm/cloud/sdk/core/util/discriminator/DiscriminatorBasedTypeAdapterFactoryTest.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Test
+  public void deserializedWithoutDiscriminatorMappingShouldResultTheSuppliedType() {
+    String json = "{ \"action\":\"accept\"}";
+    Object result = gson.fromJson(json, StatusPayloadWithoutDiscriminatorMapping.class);
+    assertEquals(result.getClass().getName(), StatusPayloadWithoutDiscriminatorMapping.class.getName());
+    assertNotEquals(result.getClass().getName(), AcceptPayloadWithoutDiscriminatorMapping.class.getName());
+  }
+
```
pyrooka, line 46:
```suggestion
```

src/test/java/com/ibm/cloud/sdk/core/util/discriminator/ResolvePayload.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.ibm.cloud.sdk.core.util.discriminator;
+
+public class ResolvePayload extends StatusPayload {
+
+}
```
pyrooka, line 17: Same as in the first 2 change suggestions.
```suggestion
public class ResolvePayload extends StatusPayload {
}
```

src/test/java/com/ibm/cloud/sdk/core/util/discriminator/StatusPayload.java
```diff
@@ -0,0 +1,36 @@
...
+
+  protected String action;
+
+  protected StatusPayload() {
+  }
+
+  public String action() {
+    return action;
+  }
+
```
pyrooka, line 34:
```suggestion
```

src/test/java/com/ibm/cloud/sdk/core/util/discriminator/StatusPayloadWithoutDiscriminatorMapping.java
```diff
@@ -0,0 +1,39 @@
...
+
+  protected String action;
+
+  protected StatusPayloadWithoutDiscriminatorMapping() {
+  }
+
+  public String action() {
+    return action;
+  }
+
```
pyrooka, line 37:
```suggestion
```

src/test/java/com/ibm/cloud/sdk/core/util/discriminator/AcceptPayloadWithoutDiscriminatorMapping.java
pyrooka: I would go without newline, but both version is ok (according to the Google style guide), so it's up to you which one you choose.
```suggestion
public class AcceptPayloadWithoutDiscriminatorMapping extends StatusPayloadWithoutDiscriminatorMapping {}
```

pyrooka (approved):
Some minor code style suggestion, but it looks good overall.

Andras-Csanyi:
Let's hold off on this PR for a while.

padamstx:
@Andras-Csanyi I think we can go ahead and change pom.xml to use 1.8 in the compiler plugin config.  The project requires a min level of java 8.  I think the pom.xml setting was overlooked when we moved up to java 8 as the min level.

padamstx:
One other comment while I'm reviewing... I've studied the new testcase along with the new model classes (StatusPayload, etc.), and I'm curious how these additional tests have caused the coverage to increase.   In the `deserializeShouldProduceTheProperObjectType` test method, I don't understand how that is testing anything that is not already tested by the DiscriminatorSerializationTest class.   Also, the other two test methods (one with a missing discriminator mapping and the other with a missing discriminator property name) are testing the unmarshaling of models which do not contain a discriminator.
So just curious how the test coverage results would have improved for any of these three test methods :)

Andras-Csanyi:
> One other comment while I'm reviewing... I've studied the new testcase along with the new model classes (StatusPayload, etc.), and I'm curious how these additional tests have caused the coverage to increase. In the `deserializeShouldProduceTheProperObjectType` test method, I don't understand how that is testing anything that is not already tested by the DiscriminatorSerializationTest class. Also, the other two test methods (one with a missing discriminator mapping and the other with a missing discriminator property name) are testing the unmarshaling of models which do not contain a discriminator.
> So just curious how the test coverage results would have improved for any of these three test methods :)
> 
> Edit: I compared the test coverage from the main branch with the coverage numbers for your feature branch and they seem the same to me. Let's discuss.

It is due to IntelliJ. When you tell IntelliJ `Run all tests` with coverage you have to select whether the tests are `jUnit` or `TestNG`. I selected junit because I saw we use that in this project and the coverage result showed that this discriminator is not covered because only jUnit tests were executed. However, in reality these tests are using `TestNG`. But, I didn't look at the original Jacoco report, it was my mistake to trust what IntelliJ shows me and not double check things, which merges coverage from junit and testng execution due to that `maven test` executes both (this is also something I have to double check and have a solid understanding). That is my understanding right now. I need to dig deeper in `what IntelliJ does?` topic to understand better.

Based on the short discussion we had on this topic yesterday it is clear for me that jacoco report produced by `maven test` is the single source of truth and IntelliJ, or any tool we use, is just a complementary stuff. Moreover, if I'm not mistaken, Codecov and Sonar processes the jacoco report.

padamstx:
> Based on the short discussion we had on this topic yesterday it is clear for me that jacoco report produced by `maven test` is the single source of truth

Yes, I think most of our "official" metrics should come from a command-line build using the "mvn" command, similar to how our automated travis builds are done.
On a slightly different point... I would be in favor of us changing our java core JUnit-based tests to use TestNG instead, so that we're consistent across the board.  It should be a simple matter of changing a few imports for various annotations and possibly a few coding changes to deal with the differences associated with setup methods, etc.  Could you please open an issue for this work?

Andras-Csanyi:
> > Based on the short discussion we had on this topic yesterday it is clear for me that jacoco report produced by `maven test` is the single source of truth
> 
> Yes, I think most of our "official" metrics should come from a command-line build using the "mvn" command, similar to how our automated travis builds are done.
> On a slightly different point... I would be in favor of us changing our java core JUnit-based tests to use TestNG instead, so that we're consistent across the board. It should be a simple matter of changing a few imports for various annotations and possibly a few coding changes to deal with the differences associated with setup methods, etc. Could you please open an issue for this work?

I did it a few hours ago.

PR #131, opened 2021-07-08
fix: bump maven-compiler-plugin source and target version to 1.8

Andras-Csanyi:
This PR increases the `source` and `target` version of maven-compiler-plugin to `1.8`.

The whole platform SDK test set was executed on my local machine using a snapshot build, including this change, from the core and all tests passed.


padamstx (approved):
LGTM

PR #132, opened 2021-07-12
feat(BaseService): remove final modifier from createServiceCall

ricellis:
Remove the `final` modifier from the `BaseService#createServiceCall` to allow extending services to override.

This provides a hook point that can be used to modify every request earlier than an interceptor is able to do so.

Added a test that will fail to compile if the final modifier is re-added inadvertently. The assertion itself is slightly pointless as it effectively validates the JVM's override functionality, but it seems nicer than having a "compile-only" test.

Note also as per [JLS 13.4.17](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.17) is a binary compatible change.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/service/BaseService.java
@@ -322,7 +322,7 @@ public abstract class BaseService {
    * @param converter the converter
    * @return the service call
    */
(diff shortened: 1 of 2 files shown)


padamstx (approved):
LGTM
Thank you for adding the test, and for including the reference to the info re: binary compatibility!

PR #134, opened 2021-07-12
chore(tests): remove junit completely

Andras-Csanyi:
By this PR:
- `junit` is removed completely and `TestNG` is used and
- there were a few custom exception checks not utilizing the testing framework exception checks and
- `maven-surefire-plugin` uses forks for execution (1 fork) which is not reused and 
- mockito and powermock versions are bumped to the most recent version


padamstx (approved):
LGTM!

dpopp07 (approved):
I don't know the Java Core well, but this looks good as far as I can tell 👍

PR #135, opened 2021-07-18
test: unit tests covering LongToDateTypeAdapter

Andras-Csanyi:
This Pr brings 100% coverage for the `LongToDateTypeAdapter`. This adapter is not used in the Gson setup the core provides by default, but part of the core and has to be covered by tests.

The overall coverage looks like below.

Note the following:

- I went a little bit further than 100% coverage and created a few extra tests simulating real world cases
- cases where deprecated constructors of Date class used, like `new Date(1990, 1, 1)` aren't covered
- there is a discrepancy of how the case is managed when empty string the value in the json string ( `{date:""}`, this adapter throws exception, but the `BooleanToStringTypeAdapter` handles it as `NULL` value. I provided a test covering this for documentation pusposes and created a ticket on our board about to check consistency across TypeAdapters around this case

Screenshot 2021-07-18 at 13 02 28


padamstx (approved):
LGTM

dpopp07 (approved):
Looks good to me 👍

PR #150, opened 2021-10-05
fix: Prevent double entries of Authorization header on second authenticate call

eiri:
When `authenticate` method called by `BasicAuthenticator`, `BearerTokenAuthenticator` or `TokenRequestBasedAuthenticator` second time on the same request it adds second `Authorization` header to the request. This can lead to a consecutive failure with `400` error from the services with strict rules about headers' double entries.

This PR switches named authenticators to use `builder.header` method to prevent that.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "package-lock.json|^.secrets.baseline$",
     "lines": null
   },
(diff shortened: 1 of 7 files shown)


padamstx (approved):
Changes look good!

PR #153, opened 2021-10-12
feat(VpcInstanceAuthenticator): add support for new VPC authentication flow

padamstx:
This commit introduces the VpcInstanceAuthenticator.
This authenticator implements the authentication flow
within a VPC-managed compute resource that is configured to
use the compute resource identity feature.
This involves the use of the compute resource's local
VPC Instance Metadata Service API to retrieve an instance identity
token, and then exchange that token for an IAM access token.
The IAM access token is then used to authenticate outbound REST
API requests by adding an Authorization containing the access token.


padamstx:
@dpopp07 @pyrooka I've pushed a commit to this PR that refactors the VpcInstanceAuthenticator class so that it does in fact subclass the TokenRequestBasedAuthenticator base class, similar to the CP4D authenticator.
Our webex session today opened my eyes a bit and I realized that I could use the CP4D pattern here without any problems.
The result is that I was able to remove some of the code within VpcInstanceAuthenticator, and avoid some duplication.

PR #158, opened 2021-12-01
fix: update logback version to fix vulnerability

rmkeezer:
Logback core module recently added a `1.2.7` version that fixes an existing vulnerability. This updates to that new version.


padamstx (approved):
LGTM

PR #164, opened 2022-02-03
build: refine criteria for semrel build stage

padamstx:
This commit removes the "type = push" criteria from the semrel build stage so that we can manually trigger a build of the main branch and also cause the semrel stage to fire.   This was needed because for a manually triggered build, the build type is "api" and not "push".


rmkeezer (approved):
LGTM

PR #165, opened 2022-02-18
fix: set visibility of ConfigBasedAuthenticatorFactory.createAuthenticator to protected

padamstx:
This commit changes the visibility of the
"ConfigBasedAuthenticatorFactory.createAuthenticator()" method from
private to protected to allow SDKs to override this class
in order to support alternate configuration sources
(e.g. injected spring properties).


tomblench (approved):

dpopp07 (approved):
Looks good! 👍

PR #166, opened 2022-02-22
chore: ignore kotlin-stdlib vuln false positive

padamstx:
This PR adds a .cveignore file entry for the kotlin-stdlib vulnerability which is a false positive.


rmkeezer (approved):
LGTM

PR #169, opened 2022-03-15
feat: support user-defined retry interceptors

padamstx:
This commit provides a pluggable way for users of the java core
to supply their own retry interceptor implementation in order
to customize the conditions under which retries are automatically
attempted.
Included in these changes is the notion of a "retry strategy",
which is a factory for creating retry interceptor instances.
Users of the java core (e.g. an SDK project) can implement
their own retry strategy and then set it with the
HttpClientSingleton.setRetryStrategy() static method.
Once set, the registered factory's createRetryInterceptor()
method will be called whenever the java core needs to initialize an
http client instance when retries are enabled.
The factory's createRetryInterceptor() method returns an implementation
of the IRetryInterceptor interface (which is itself a subclass of
the okhttp3 Interceptor interface).
User's might find it convenient to define their retry interceptor
class as a subclass of the java core's RetryInterceptor class,
thereby relying on RetryInterceptor for most of the functionality,
while overriding specific methods as needed.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
@@ -15,7 +15,6 @@ package com.ibm.cloud.sdk.core.http;
 
 import com.ibm.cloud.sdk.core.http.HttpConfigOptions.LoggingLevel;
 import com.ibm.cloud.sdk.core.http.gzip.GzipRequestInterceptor;
-import com.ibm.cloud.sdk.core.service.BaseService;
 import com.ibm.cloud.sdk.core.service.security.DelegatingSSLSocketFactory;
 import okhttp3.Authenticator;
 import okhttp3.ConnectionSpec;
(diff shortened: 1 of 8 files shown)


src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
```diff
@@ -320,9 +333,33 @@ private OkHttpClient reconfigureClientInterceptors(OkHttpClient client, String i
...
+          iter.remove();
+        }
+      }
+    }
+
+    return builder.build();
+  }
+
+  /**
+   * Sets a new list of interceptors for the specified {@link OkHttpClient} instance by removing any interceptors
```
padamstx, line 347: This new flavor of reconfigureClientInterceptors() was needed because we can no longer key on the class name of the interceptor class that we want to remove.   Instead, we'll remove any interceptor that implements the IRetryInterceptor marker interface.

src/main/java/com/ibm/cloud/sdk/core/http/RetryInterceptor.java
```diff
@@ -103,8 +133,13 @@ public Response intercept(Interceptor.Chain chain) throws IOException {
...
 
-  // Get the time we should wait before fire the next request in milliseconds.
-  private int getInterval(Response response, Request request) {
+  /**
+   * Determine the retry interval to wait before attempting the next retry.
+   * @param response the response from the previously attempted request
+   * @param request the previously attempted request
+   * @return the retry interval in milliseconds
+   */
+  protected int getInterval(Response response, Request request) {
```
padamstx, line 142: The primary methods in this class were changed from private to protected to enable users to subclass this implementation.  I think the primary use case for user-defined retry interceptors will probably be to subclass
this class and then override the "shouldRetry()" method below.

padamstx:
@richellis Rich, just wanted to alert you to this PR in case you have any comments.

dpopp07 (approved):
Looks good to me! 👍

PR #170, opened 2022-03-16
fix: bump gson to 2.9.0

padamstx:
This PR bumps our gson dependency to the current latest (2.9.0).
Included in this are changes to wean us off the GSON ReflectionAccessor class, which is no longer present in v2.9.0.


dpopp07 (approved):
Looks good to me 👍

pyrooka (approved):
LGTM!

PR #174, opened 2022-04-01
feat: allow user-specified default OkHttpClient instance

padamstx:
This commit modifies the java core so that users can now
specify a fully-configured OkHttpClient instance to be
used by service and authenticator instances:
1. The HttpClientSingleton class now has methods setHttpClient()
   and getHttpClient().  These methods can be used to set and get
   the default client configuration from which other OkHttpClient
   instances are created.
2. The TokenRequestBasedAuthenticator base class now has methods
   setClient() and getClient().  These methods can be used to set
   and get the specific OkHttpClient instance that should be used
   by an authenticator when interacting with the token service.
   These new methods are inherited by each of the request-based
   authenticator implementation classes: ContainerAuthenticator,
   CloudPakForDataAuthenticator, CloudPakForDataServiceAuthenticator,
   CloudPakForDataServiceInstanceAuthenticator, IamAuthenticator, and
   VpcInstanceAuthenticator.

Note: ~~I plan to update this PR slightly so that the HttpClientSingleton will create a default OkHttpClient instance that supports  mutual TLS (if possible).  First, I need to learn what those details are :)   The PR can be reviewed in its current state as these follow-on changes should be fairly small and isolated to the HttpClientSingleton class.~~
Upon further review, I'm going to go ahead with this PR as-is.  If additional research into the mTLS topic reveals a path where I can configure the default OkHttpClient instance such that mTLS works out of the box without affecting existing use-cases, then I will follow-up with a new PR to make those changes.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2019, 2021.
+ * (C) Copyright IBM Corp. 2019, 2022.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


padamstx:
@ricellis just wanted to alert you to this PR in case you have any comments or concerns.  Also, if you have any ideas about how an OkHttpClient instance needs to be configured to support mutual TLS authentication between the client and server, let me know.

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
```diff
@@ -228,7 +273,7 @@ public void run() {
         }
       }
 
-    // Return the access token from our stored tokenData object.
+    // Return the access token from our stored tokenData object.object
```
pyrooka, line 276: I guess this wasn't intentional.

pyrooka (approved):
LGTM!

src/main/java/com/ibm/cloud/sdk/core/security/TokenRequestBasedAuthenticator.java
padamstx: good catch!

dpopp07 (approved):
Looks good to me! 👍

PR #175, opened 2022-04-05
fix: support JSSE system properties

padamstx:
This commit makes a small change to the
HttpClientSingleton.setupTLSProtocol() method
to allow for the use of mutual TLS (i.e. client
authentication).
Specifically, the code changes will allow the JSSE-related
system properties to be used to specify a client truststore
and/or keystore:
- javax.net.ssl.trustStore
- javax.net.ssl.trustStorePassword
- javax.net.ssl.keyStore
- javax.net.ssl.keyStorePassword

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/HttpClientSingleton.java
@@ -117,11 +117,13 @@ public class HttpClientSingleton {
         // Find the TLS protocols supported by this socket
         List<String> supportedTlsNames = Arrays.asList(socket.getSupportedProtocols());
         LOG.log(Level.FINEST, "Socket supported protocols {0}", supportedTlsNames);
+
         // Get the union of MODERN_TLS_NAMES and the socket's supported protocols
         List<String> protocolsToEnable = new ArrayList<>();
         protocolsToEnable.addAll(supportedTlsNames);
         protocolsToEnable.retainAll(MODERN_TLS_NAMES);
         LOG.log(Level.FINEST, "Filtered protocols to enable {0}", protocolsToEnable);
+
         socket.setEnabledProtocols(protocolsToEnable.toArray(new String[]{}));
         return socket;
       }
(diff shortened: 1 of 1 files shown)


ricellis:
Isn't this a breaking change?
Consider a case where the _default_ `SSLContext` for whatever reason doesn't enable the `TLS` protocol. The [existing code using `getInstance("TLS")`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLContext.html#getInstance-java.lang.String-) will go through the list of configured providers looking for one that does support `TLS`, but this new code will not and will fail instead; ulimately forcing a user configuration change to add TLS support in their default context (which is presumably not desirable if it hasn't been configured as such) or to otherwise customize the `SSLSocketFactory`.

I'm not sure just how _significant_ that risk is, but nonetheless I feel it is a breaking behaviour change and it probably disproportionately impacts server environments with different frontend/backend contexts.

I also have concerns about ` disableSslVerification` (beyond my usual ones about how I dislike that option 😄) - it won't delegate to the default `SSLContext` instead creating one in the current style, so there will be mismatch in behaviour depending if that toggle is on (beyond the existing one I've just noticed where `SSL` and `TLS` may not be necessarily be provided by the same provider anyway) .

Looking at this are of code in general I think we should probably consider the balance of the most common use case vs where customization is required by the user before making a breaking change. For example using `cacerts` by default is more likely to succeed when connecting to the IBM Cloud services with public certs than using _any other_ configured trust store. Further when running the SDKs in e.g. an app in a WAS Liberty runtime then there are some [specific caveats around what is honoured and what isn't](https://www.ibm.com/docs/en/was-liberty/base?topic=liberty-ssl-defaults-in) and expectations for third-party (i.e. non-WAS) code to call `SSLSocketFactory.getDefault()` rather than `SSLContext.getDefault()`.

In the meantime #174 provides an option for customizing right? via something like:
```java
getHttpClient().newBuilder().sslSocketFactory(...).build()
```

padamstx:
@ricellis and I discussed this via slack and decided not to go with these changes due to the potential change in default behavior for users.   It's not clear how popular mTLS is among our users, so we need to be careful not to potentially break existing (non-mTLS) users in order to solve a problem that has unknown scope/prevalence. I'll be looking at alternatives and will submit a new PR if/when appropriate.

PR #177, opened 2022-04-28
chore: improve retry conditions

pyrooka:
This commit adds an extra status code check to the `shouldRetry` method.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/http/RetryInterceptor.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/http/RetryInterceptor.java
@@ -186,7 +186,7 @@ public class RetryInterceptor implements IRetryInterceptor {
    */
   protected boolean shouldRetry(Response response, Request request) {
     // First check the response.
-    if (response.code() == 429 || (response.code() >= 500 && response.code() != 501)) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/ibm/cloud/sdk/core/http/RetryInterceptor.java
```diff
@@ -186,7 +186,7 @@ protected int getInterval(Response response, Request request) {
    */
   protected boolean shouldRetry(Response response, Request request) {
     // First check the response.
-    if (response.code() == 429 || (response.code() >= 500 && response.code() != 501)) {
+    if (response.code() == 429 || (response.code() >= 500 && response.code() <= 599 && response.code() != 501)) {
```
padamstx, line 189: I don't think this change has any material effect on things... i don't think we'd expect to ever get a status code > 599 :)
But if we do, then this change nails it :joy:

padamstx (approved):
LGTM

src/main/java/com/ibm/cloud/sdk/core/http/RetryInterceptor.java
pyrooka: I know, but we have the exactly same check in the [Go core](https://github.com/IBM/go-sdk-core/blob/main/v5/core/base_service.go#L789) as well and I just wanted to get it aligned.

PR #187, opened 2022-11-10
fix: use gson 'lazily parsed number' for 'object to number' strategy

padamstx:
This commit makes a small change to our Gson factory so that it sets the "object to number" strategy to LAZILY_PARSED_NUMBER. The default value (DOUBLE) does not provide the fidelity needed when performing marshalling/unmarshalling steps involving generic datastructures like Map<String,Object>.

Signed-off-by: Phil Adams <phil_adams@us.ibm.com>

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/util/GsonSingleton.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/util/GsonSingleton.java
@@ -47,7 +47,7 @@ public final class GsonSingleton {
    */
   private static Gson createGson(boolean prettyPrint, boolean serializeNulls) {
     GsonBuilder builder = new GsonBuilder()
-        .setObjectToNumberStrategy(ToNumberPolicy.DOUBLE)
+        .setObjectToNumberStrategy(ToNumberPolicy.LAZILY_PARSED_NUMBER)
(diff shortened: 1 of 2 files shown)


ricellis (commented):
This is the same strategy already used in the untyped `DynamicModel<Object>` (which IIRC are also kept in `Map<String, Object>`) - I think it is the only way to retain number fidelity using GSON when the actual type isn't known. What I don't really understand is where this fails, because if types are loose then shouldn't they be declared as schemas that end up as `DynamicModel<Object>` and go down the path already using the `LAZILY_PARSED_NUMBER` strategy?

I'm not opposed, but I think it is worth noting that this is potentially a breaking change. I don't know how widespread usage of `Map<String, Object>` types are in consuming SDKs, but anyone assuming the current behaviour could be doing things like:
`double d = (double) map.get("foo");`
after this change that will fail with
`java.lang.ClassCastException: com.google.gson.internal.LazilyParsedNumber incompatible with java.lang.Double`
and will need to be updated to:
`double d = ((Number) map.get("foo")).doubleValue();`

padamstx:
Discussed this with @RICELLIS on slack and we agreed that moving forward with this change should be ok (famous last words perhaps :joy: ).

pyrooka (approved):
LGTM!

PR #196, opened 2023-01-20
refactor: remove guava dependency

ricellis:
The guava dependency is large and causes [CVE-2020-8908](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908) to get flagged by some scanning tools (though to be clear the vulnerable guava method is not used in this code).
Since guava is used in just a few places it could be removed to reduce the size of the SDKs and prevent this false positive vulnerability warning.

This PR:
* Removes `guava-version` and the guava dependency block from the pom
* Updates `RequestBuilder` to use okhttp's path segment encoding instead of guava's URL escaper
* Replaces guava base64 encoding/decoding with `java.util.Base64`
* Replaces some list/set initialization in test classes.
* Finally, removes `javax.annotation.Nullable` (as supplied by the guava package) that was referenced by `HttpConfigTest`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,6 @@
         <jacoco-plugin-version>0.8.8</jacoco-plugin-version>
         <compiler-plugin-version>3.10.1</compiler-plugin-version>
         <okhttp3-version>4.10.0</okhttp3-version>
-        <guava-version>30.1.1-jre</guava-version>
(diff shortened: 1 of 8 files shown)


padamstx (approved):
LGTM

padamstx:
@ricellis Thanks for making this change!

PR #200, opened 2023-02-28
fix: recognize content types with whitespace

dpopp07:
The RFC for content-type header format allows optional whitespace before the semicolon. Our regular expressions for recognizing various mime types do not account for this, so valid mime types may not be recognized. This commit adjusts the expression to close that gap.


padamstx (approved):
LGTM

PR #202, opened 2023-05-03
fix(ContainerAuthenticator): add sa-token as default CR token filename

padamstx:
This commit adds support for a second default CR token filename. If the user doesn't specify a CR token filename, the authenticator will first try '/var/run/secrets/tokens/vault-token', and then '/var/run/secrets/tokens/sa-token' in that order.


pyrooka (approved):
LGTM!

dpopp07 (approved):
Looks good and confirmed with internal cluster 👍

PR #206, opened 2023-05-23
fix(RequestBuilder): retain trailing slash on operation path

padamstx:
This PR fixes RequestBuilder.resolveRequestUrl() so that it will retain a trailing slash found on an operation path when adding the path segments to form the request URL.


dpopp07 (approved):
Looks good! 👍

PR #208, opened 2023-07-14
build: remove 'latest' link from gh-pages

padamstx:
This commit modifies the publishJavadoc.sh script
to remove the creation of the "latest" symbolic
link.  This is being done because github no longer allows us to include symbolic links in the gh-pages branch.


dpopp07 (approved):
👍

PR #216, opened 2023-10-17
Upgrade okhttp 4.12.0

ricellis:
Upgrade okhttp to 4.12.0.
Revert the okio version override since 4.12.0 has a newer dependency.


padamstx (approved):
LGTM

PR #221, opened 2024-02-23
fix: adjust IAM token expiration time

padamstx:
This commit changes the IAM, Container and VPC Instance authenticators slightly so that an IAM access token will be viewed as "expired" when the current time is within 10 seconds of the official expiration time. IOW, we'll expire the access token 10 secs earlier than the IAM server-computed expiration time.
We're doing this to avoid a scenario where
an IBM Cloud service receives a request along
with an "almost expired" access token and then uses that token to perform downstream requests in a
somewhat longer-running transaction and then the
access token expires while that transaction is
still active.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
@@ -1,5 +1,5 @@
 /**
- * (C) Copyright IBM Corp. 2015, 2021.
+ * (C) Copyright IBM Corp. 2015, 2024.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
  * the License. You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
```diff
@@ -119,11 +127,18 @@ public synchronized boolean needsRefresh() {
...
+   * @return true iff the current access token is valid and not expired
    */
   @Override
   public boolean isTokenValid() {
     return this.getException() == null
-        && (this.expiration == null || Clock.getCurrentTimeInSeconds() < this.expiration);
+        && (this.expiration == null || Clock.getCurrentTimeInSeconds() < (this.expiration - IamExpirationWindow));
+  }
+
+  public String toString() {
```
padamstx, line 138: This was useful during debugging, so I thought I'd leave it here.

src/test/resources/iam_token.json
```diff
@@ -3,5 +3,5 @@
   "refresh_token": "00000000",
   "token_type": "Bearer",
   "expires_in": 3600,
-  "expiration": 1522788645
+  "expiration": 1600000000
```
padamstx, line 6: Debugging is easier with round numbers :joy:

src/main/java/com/ibm/cloud/sdk/core/security/IamToken.java
pyrooka: I was wondering if it's a good idea to automatically print the token by calling a single method, but I guess it's fine since the method belongs to the `IamToken` class so it should be obvious what it does.

pyrooka (approved):
Looks good!

dpopp07 (approved):
Looks good! 👍

PR #245, opened 2026-04-16
build: bump java build version to 25

Andris28:
In this commit we bump the
java build version from 17 to 25.


pyrooka (approved):
Looks good!

PR #247, opened 2026-06-04
feat(auth): add support for new service version in VPC Instance Auth

diatrcz:
This PR adds support for the new VPC Instance Metadata Service version (2025-08-26) with the following changes:

- Added configurable `ServiceVersion` field (default: "2022-03-01") that switches to new API paths (`/identity/v1/token`, `/identity/v1/iam_tokens`) when set to "2025-08-26"
- Added `TokenLifetime` field (default: 300s) for customizable token expiration
- Configuration via builder methods (`SetServiceVersion()`, `SetTokenLifetime()`) or `VPC_IMS_VERSION` environment variable
- **Full backward compatibility with older service versions**


src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
```diff
@@ -121,6 +133,27 @@ public Builder url(String url) {
...
+    public Builder serviceVersion(String serviceVersion) {
+      this.serviceVersion = serviceVersion;
+      return this;
+    }
+
+    /**
+     * Sets the tokenLifetime Property.
+     * @param tokenLifetime the base token lifetime to use.
+     * @return the Builder
+     */
```
pyrooka, line 155: 🫣
```suggestion
    /**
     * Sets the tokenLifetime Property.
     * 
     * @param tokenLifetime the base token lifetime to use.
     * @return the Builder
     */
```

src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
```diff
@@ -236,10 +277,71 @@ protected void setURL(String url) {
     this.url = url;
   }
 
+  /**
+   * @return the VPC ServiceVersion configured in this Authenticator.
```
pyrooka: Just a little change would be nice to clarify that the version is not for the VPC service but the VPC IMS.
```suggestion
   * @return the VPC Instance Metadata Service ServiceVersion configured in this Authenticator.
```

src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
```diff
@@ -236,10 +277,71 @@ protected void setURL(String url) {
...
+   */
+  public String getServiceVersion() {
+    return this.serviceVersion;
+  }
+
+    /**
+   * Sets the ServiceVersion in this Authenticator.
+   *
+   * @return the VPC ServiceVersion
```
pyrooka, line 293: Same as above.

src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
```diff
@@ -236,10 +277,71 @@ protected void setURL(String url) {
...
     return (StringUtils.isEmpty(this.url) ? defaultIMSEndpoint : this.url);
   }
 
+  /**
+   * Gets the operation path for creating an access token based on the service version.
+   *
+   * @return the correct access token path
+   */
+  public String getCreateAccessTokenPath() {
+    if (this.serviceVersion.equals("2025-08-26")) {
```
pyrooka: I know I initially implemented this the same way, but since this value is used multiple times in the code, it might be worth assigning it to a var/const. What do you think?

pyrooka (changes requested):
Left a few minor suggestions/question, but looks pretty good overall!

pyrooka (approved):
LGTM!

PR #250, opened 2026-07-07
feat(auth): add new IAM profile selector to VPC Instance Auth

diatrcz:
Adds IAM Profile selector to the `VPC Instance Authenticator`.

- The name of the linked trusted IAM profile to be used as the identity of the compute resource.
- At most one of IAMProfileCRN, IAMProfileID, or IAMProfileName may be specified.
- If none is specified, the default IAM profile of the compute resource will be used.

### DIFF
--- a/src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
@@ -480,7 +480,7 @@ public class VpcInstanceAuthenticator
         requestBody = String.format("{\"trusted_profile\": {\"id\": \"%s\"}}", getIamProfileId());
       }
       if (!StringUtils.isEmpty(getIamProfileName())) {
-        requestBody = String.format("{\"trusted_profile\": {\"id\": \"%s\"}}", getIamProfileName());
+        requestBody = String.format("{\"trusted_profile\": {\"name\": \"%s\"}}", getIamProfileName());
       }
 
       // If we created a request body above, then set it on the request now.
--- a/src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
+++ b/src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
@@ -58,6 +58,7 @@ public class VpcInstanceAuthenticator
   // Properties specific to a VpcInstanceAuthenticator.
   private String iamProfileCrn;
   private String iamProfileId;
+  private String iamProfileName;
   private String url;
   private String serviceVersion;
   private int tokenLifetime;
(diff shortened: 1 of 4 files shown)


src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
```diff
@@ -208,10 +228,21 @@ public static VpcInstanceAuthenticator fromConfiguration(Map<String, String> con
...
-    if (StringUtils.isNotEmpty(getIamProfileCrn()) && StringUtils.isNotEmpty(getIamProfileId())) {
+    // At most one of iamProfileCrn or iamProfileI or iamProfileName may be specified.
+    int counter = 0;
+    if (StringUtils.isNotEmpty(getIamProfileCrn())) {
+      counter++;
+    }
+    if (StringUtils.isNotEmpty(getIamProfileId())) {
+      counter++;
+    }
+    if (StringUtils.isNoneEmpty(getIamProfileName())) {
```
pyrooka: I think we should stick to the other function that we already use :)
```suggestion
    if (StringUtils.isNotEmpty(getIamProfileName())) {
```

src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
```diff
@@ -208,10 +228,21 @@ public static VpcInstanceAuthenticator fromConfiguration(Map<String, String> con
    */
   @Override
   public void validate() {
-    // At most one of iamProfileCrn or iamProfileId may be specified.
-    if (StringUtils.isNotEmpty(getIamProfileCrn()) && StringUtils.isNotEmpty(getIamProfileId())) {
+    // At most one of iamProfileCrn or iamProfileI or iamProfileName may be specified.
```
pyrooka: 👀 
```suggestion
    // At most one of iamProfileCrn or iamProfileId or iamProfileName may be specified.
```

pyrooka (changes requested):
The PR looks good, but we should address the two minor issues from my review before merging.

src/main/java/com/ibm/cloud/sdk/core/security/VpcInstanceAuthenticator.java
```diff
@@ -208,10 +228,21 @@ public static VpcInstanceAuthenticator fromConfiguration(Map<String, String> con
...
+      counter++;
+    }
+    if (StringUtils.isNoneEmpty(getIamProfileName())) {
+      counter++;
+    }
+
+    if (counter > 1) {
       throw new IllegalArgumentException(
-          String.format(ERRORMSG_ATMOST_ONE_PROP_ERROR, "iamProfileCrn", "iamProfileId"));
+          String.format(ERRORMSG_ATMOST_ONE_PROP_ERROR, "iamProfileCrn", "iamProfileId", "iamProfileName"));
```
pyrooka, line 245: One more thing: the `ERRORMSG_ATMOST_ONE_PROP_ERROR` constant should be updated to handle 3 arguments. Previously it was just 2.

pyrooka (approved):
LGTM!

